PR #2, opened 2016-09-06
David Davydov - Week 3 Questions

kctrlv:
This exercise made me feel like there's a lot I don't fully know, but I know a lot more than I did a week ago. There's a lot of material in this course, and it helps to keep up with all that we've covered during the week.

### DIFF
--- a/module-2/backend/week_three.md
+++ b/module-2/backend/week_three.md
@@ -10,16 +10,68 @@ Note: When you're done, submit a PR with a reflection in the comments about how 
 ### Questions
 
 1. What are the 3 main features in an ERD?
+```
+Entities: Objects that we store the information about.
+Relationships: The way each entities are linked to one another
+Cardinalities: An attribute of each relationship, can either be "One to One", "One to Many", or "Many to Many"
+```
 2. Create an ERD for the following database: Restaurants, Customers, Items, Ingredients, Beverages, Orders, Vendors.
+```
+
+```
 3. What is the entry at the command line to create a new rails app?
+```
+rails new app_name
+```
 4. What do Models generally inherit from in rails?
+```
+Models inherit from ActiveRecord::Base
+```
 5. What do Controllers generally inherit from in a rails project?
+```
+Controllers inherit from ActionController::Base
+```
 6. How would I create a route if I wanted to see a specific horse in my routes fitle assuming I'm sticking to standard conventions and that I didn't want other CRUD functionality?
+```
+resources :horses, only [:show]
+```
 7. What rake task is useful when looking at routes, and what information does it give you?
+```
+'rake routes' prints the list of routes that have been defined in the project's routes file.
+```
 8. What is an example of a route helper? When would you use them?
+```
+horses_path is equivalent to '/horses' and horse_path(id) is the same as '/horses/:id'
+Route helpers are useful because often times we do not know the exact url or pathname that we want to access, whereas that information is dynamically accessible through the route helper.
+```
 9. What's the difference between what `_url` and `_path` return when combined with a routes prefix?
+```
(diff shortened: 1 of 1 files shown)


module-2/backend/week_three.md
```diff
@@ -10,16 +10,68 @@ Note: When you're done, submit a PR with a reflection in the comments about how
 ### Questions
 
 1. What are the 3 main features in an ERD?
+```
+Entities: Objects that we store the information about.
+Relationships: The way each entities are linked to one another
+Cardinalities: An attribute of each relationship, can either be "One to One", "One to Many", or "Many to Many"
+```
 2. Create an ERD for the following database: Restaurants, Customers, Items, Ingredients, Beverages, Orders, Vendors.
+```
```
case-eee: did you create this ERD? if so, do you have questions about it?

module-2/backend/week_three.md
```diff
@@ -10,16 +10,68 @@ Note: When you're done, submit a PR with a reflection in the comments about how
...
+rails new app_name
+```
 4. What do Models generally inherit from in rails?
+```
+Models inherit from ActiveRecord::Base
+```
 5. What do Controllers generally inherit from in a rails project?
+```
+Controllers inherit from ActionController::Base
```
case-eee: controllers generally inherit from just `ApplicationController`

module-2/backend/week_three.md
```diff
@@ -10,16 +10,68 @@ Note: When you're done, submit a PR with a reflection in the comments about how
...
+```
 6. How would I create a route if I wanted to see a specific horse in my routes fitle assuming I'm sticking to standard conventions and that I didn't want other CRUD functionality?
+```
+resources :horses, only [:show]
```
case-eee: how could you create this without using `resources`?

module-2/backend/week_three.md
```diff
@@ -10,16 +10,68 @@ Note: When you're done, submit a PR with a reflection in the comments about how
...
+```
 7. What rake task is useful when looking at routes, and what information does it give you?
+```
+'rake routes' prints the list of routes that have been defined in the project's routes file.
```
case-eee: what other information does it give you? it gives you the prefix, URI pattern, verb, controller & action

module-2/backend/week_three.md
```diff
@@ -10,16 +10,68 @@ Note: When you're done, submit a PR with a reflection in the comments about how
...
+```
 8. What is an example of a route helper? When would you use them?
+```
+horses_path is equivalent to '/horses' and horse_path(id) is the same as '/horses/:id'
+Route helpers are useful because often times we do not know the exact url or pathname that we want to access, whereas that information is dynamically accessible through the route helper.
```
case-eee: 👍

module-2/backend/week_three.md
```diff
@@ -10,16 +10,68 @@ Note: When you're done, submit a PR with a reflection in the comments about how
...
+strong params are a way to explicitly require only specific params. They are necessary because a user might inject a parameter of their own into a POST/PUT request where we do not want there to be one, and we do not want a chance that our code will be affected by that injected parameter. By specifying only the params we expect to come back to us, we keep the application safe from such vulnerabilities.
+```
 11. What role does `form_for` play in helping us create our forms?
+```
+form_for(Object) is an easy way to create a form that fills in attributes for a particular object.
+```
+
 12. How does `form_for` know where to submit the user's input?
+```
+The controller is responsible for handling the user's input from a "form_for", likely in a POST/PUT method.
```
case-eee: how does `form_for` know to send the request to the `create` vs `update` action?

case-eee:
hey @kctrlv! left ya some comments and questions to think about. overall, good work on this! let me know if you have questions 👍

PR #5, opened 2016-09-06
my responses

bcgoss:
Answer for question 2: 

screen shot 2016-09-06 at 12 19 19 am


bcgoss:
Beverages are strange. They can go on an order. They have recipes. I would just roll them into "Items." Another strategy is to make a "menu" structure and add both items and beverages to it.

PR #6, opened 2016-09-06
Workman Update week_three.md

rdavid1099:
I can't believe how much I have learned in one week.  I was a little weary going into the questions, but I found myself answering them with relative ease.  It still baffles me the amount of information we learn and retain while going through Turing.

PR #1, opened 2016-09-05
Susi Irwin - Week 3 recap

susiirwin:
@case-eee and @s-espinosa - here is my Week 3 recap questions. I attached a drawing of the ERD from my notebook as well. There is a link to it in the questions. Thanks!

PR #8, opened 2016-09-06
Jesse's reflection on week 3 recap (see comments)

PlanetEfficacy:
A few things I had to look up
- the Acronym ERD (#1)
- the exact syntax for creating a show route; I needed to double check the colons (#6)
- the syntax for the first line of form_for (#13)

I'm feeling good about these questions in general.

### DIFF
--- a/module-2/backend/week_3_recap.md
+++ b/module-2/backend/week_3_recap.md
@@ -1,0 +1,86 @@
+## Week Three Recap
+
+### Instructions
+Fork this repository. Answer the questions to the best of your ability.
+
+Try to answer them with limited amount of external research. These questions cover the majority of what we've learned this week.
+
+Note: When you're done, submit a PR with a reflection in the comments about how this exercise went for you.
+
+### Questions
+
+1. What are the 3 main features in an ERD?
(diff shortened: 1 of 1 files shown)


module-2/backend/week_3_recap.md
```diff
@@ -0,0 +1,86 @@
...
+
+5. What do Controllers generally inherit from in a rails project?
+
+```
+Controllers inherit from ApplicationController```
+
+6. How would I create a route if I wanted to see a specific horse in my routes file assuming I'm sticking to standard conventions and that I didn't want other CRUD functionality?
+
+```
+resources :horse, only: [:show]```
```
case-eee: how would we do this without using `resources`?

case-eee:
@PlanetEfficacy glad to hear you're feeling good about the concepts. everything looks good to me 🎉

module-2/backend/week_3_recap.md
PlanetEfficacy: @case-eee get '/horse/:id', to: 'horses#show'

PR #15, opened 2016-09-07
Raphael_Barbo - week 3 check

rsbarbo:
I had to look up the answer for question one, I remember Andrew's class but I could not remember the terminology.


rsbarbo:
@case-eee @s-espinosa Again, I am really sorry for my delay, here are the answers for the week 3 diagnoses.

module-2/backend/week_three.md
```diff
@@ -10,16 +10,82 @@ Note: When you're done, submit a PR with a reflection in the comments about how
...
+Restaurants => has_many => beverages
+Customer => has_many => orders
+```
 3. What is the entry at the command line to create a new rails app?
+```
+rails new appname
+```
 4. What do Models generally inherit from in rails?
+```
+Models inherit from Active Record
```
case-eee: models inherit from `ActiveRecord::Base`

module-2/backend/week_three.md
```diff
@@ -10,16 +10,82 @@ Note: When you're done, submit a PR with a reflection in the comments about how
...
 5. What do Controllers generally inherit from in a rails project?
+```
+Application Controller
+```
 6. How would I create a route if I wanted to see a specific horse in my routes fitle assuming I'm sticking to standard conventions and that I didn't want other CRUD functionality?
+```
+there are a few ways i can go about it, but as far as the route goes:
+
+get '/horse', to: 'horses#show'
+resources :horses, only: [:show]
```
case-eee: this and line 41 are the ways to create this route. good work creating both 👍

module-2/backend/week_three.md
```diff
@@ -10,16 +10,82 @@ Note: When you're done, submit a PR with a reflection in the comments about how
...
+```
+rake routes, and it will give you the prefix, verb, uri, controller and action
+```
 8. What is an example of a route helper? When would you use them?
+```
+user_path or @user depending how you have your code set up.
+```
 9. What's the difference between what `_url` and `_path` return when combined with a routes prefix?
+```
+path starts from the '/' root and url gives the entire path including http....
```
case-eee: nice 🎉

module-2/backend/week_three.md
```diff
@@ -10,16 +10,82 @@ Note: When you're done, submit a PR with a reflection in the comments about how
...
+```
 10. What are strong params and why are the necessary?
+```
+Its my understanding that strong params limit the visibility and accessibility of information throghout the models, and it is applied for safety reasons.
```
case-eee: it limits the key/value pairs allowed to be passed in when creating our new object

module-2/backend/week_three.md
```diff
@@ -10,16 +10,82 @@ Note: When you're done, submit a PR with a reflection in the comments about how
...
+```
 11. What role does `form_for` play in helping us create our forms?
+```
+it communicates to the application what model the form belongs to, and help us gather the proper info from the user.
```
case-eee: it also generates raw html for us so we don't have to write it ourselves

module-2/backend/week_three.md
```diff
@@ -10,16 +10,82 @@ Note: When you're done, submit a PR with a reflection in the comments about how
...
+```
 12. How does `form_for` know where to submit the user's input?
+```
+it essentially uses the controller to know what path to follow.
```
case-eee: it uses the object that we pass to it to determine where to submit the data - either `create` or `update` depending on if the object already exists in the database or not

module-2/backend/week_three.md
```diff
@@ -10,16 +10,82 @@ Note: When you're done, submit a PR with a reflection in the comments about how
...
+<%= f.text_field :username %>
+
+<%= f.label :password %>
+<%= f.text_field :password %>
+
+<%= f.submit  %>
+```
 14. Why do we want to validate our models?
+```
+to make sure all the information needed/required is included in the form submitted.
```
case-eee: we want to validate our models so we don't have bad data in our database. we don't want five people with the same email address if we are expecting it to be unique

case-eee:
@rsbarbo overall, good work on this. let me know if you have questions on my comments 🎉

PR #42, opened 2016-10-10
Anthony Ciccone Update week_one.md

antciccone:


module-2/backend/week_one.md
```diff
@@ -1,32 +1,146 @@
...
+Fork this respository. Answer the questions to the best of your ability. Try to answer them with limited amount of external research. These questions cover the majority of what we've learned this week (which is a TON!).
+
+Note: When you're done, submit a PR.
+
+List the five common HTTP verbs and what the purpose is of each verb.
+1. Get - retrieve data
+* POST - sending data to the server
+* DELETE - remove data
+* PUT - will replace/update
+* cant think of another :(
```
s-espinosa: PATCH is an alternative for PUT. Generally they are interchangeable, but sticklers will tell you that PUT replaces a full record and PATCH updates part of a record.

module-2/backend/week_one.md
```diff
@@ -1,32 +1,146 @@
...
+```
+
+In the same code block, how would I pass a local variable name with a value of Mr. Ed?
+
+
+What's the purpose of ERB?
+* Erb lets us display our data to the client. It also has the ability to read ruby code when you use special syntax
+
+Why do I need a development AND test database?
+* The test database is for test and it is destroyed right after where as a development database is for non-testing use so it is not destroyed unless given a command
```
s-espinosa: Good answer. One note: the test database is only destroyed after if we specifically make it so thats the case (e.g. by using a gem like `database_cleaner`)

module-2/backend/week_one.md
```diff
@@ -1,32 +1,146 @@
...
+
+What's responsive design?
+* Ability for your design to be sized properly on different devices with our sacrificing the functionality. It will properly fit if you adjust the size of your browser.
+
+What is CRUD and why is it important?
+* C-create R-read U-update D-delete. These give us our basic actions we can do to our database
+
+What does HTTP stand for?
+* H-hyper T-text T-i forget P-i forget
```
s-espinosa: Transfer Protocol

module-2/backend/week_one.md
```diff
@@ -1,32 +1,146 @@
...
+delete 'restaurant/id' do
+  Restaurant.destroy(params[:id])
+  redirect '/restaurants'
+  end
+```   
+This enables you to destroy an instance of restaurant
+This brings you to the show erb which only displaces that particular id or that once instance of Restaurant
+
+What's a migration?
+* migrates to our .schema. rake db:migrate
```
s-espinosa: A migration is the file that we create to store the instructions on how to create/update the database, but creating the migration (the file with the instructions) is a separate step from *running* the migration (which happens when you run `rake db:migrate`). It seems like you're aware of this since you included the command above, but just want to make sure that it takes **both** steps (creating and running) to update the DB.

module-2/backend/week_one.md
```diff
@@ -1,32 +1,146 @@
...
+When you create a migration, does it automatically modify your database?
+* it modifies the attributes
+How does a model relate to a database?
+* A model holds all the objects that that we can manipulate in our db.
```
s-espinosa: I wouldn't quite say that it "holds all the objects". It's a class that we can use to 1) interact with and query the database and 2) create instances that represent specific lines in the database.

PR #44, opened 2016-10-10
Week 2 Check for Understanding.md

AliSchlereth:
Check for understanding complete.


module-2/backend/week_one.md
```diff
@@ -4,29 +4,65 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
 
-1. List the five common HTTP verbs and what the purpose is of each verb.
-2. What is Sinatra?
-4. What is MVC?
-5. Why do we follow conventions when creating our actions/path names in our Sinatra routes?
-6. What types of variables are accessible in our view templates without explicitly passing them?
+1. List the five common HTTP verbs and what the purpose is of each verb.  
+-GET - gets something  
+-POST -change/update something  
+-PUT -create something
```
s-espinosa: PUT and POST are reversed. POST is used to create, while PUT is used to update.

module-2/backend/week_one.md
```diff
@@ -4,29 +4,65 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+  ```  
+@count = 1
 8. In the same code block, how would I pass a local variable `name` with a value of `Mr. Ed`?
-9. What's the purpose of ERB?
-10. Why do I need a development AND test database?
-11. What's responsive design?
-12. What is CRUD and why is it important?
-13. What does HTTP stand for? 
+name = "Mr. Ed"
+erb :index, locals {name} (possibly something else with name?)
```
s-espinosa: Close. This is still Ruby so think of this as passing two arguments to the `erb` method - `:index`, and the locals - `locals: {name: name}`. The first name in the locals hash is the name of the variable as you're going to use it in the view, while the second is the name of the variable that you're passing it (which you set using `name = "Mr. Ed"`.

module-2/backend/week_one.md
```diff
@@ -4,29 +4,65 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+      -GET show 
+    U -GET edit page
+      -PUT submit
+    D -DELETE
+18. What's a migration?  
+.. Something that changes the structure of the database 
+19. When you create a migration, does it automatically modify your database?  
+..  First you have to create the migration then you have to run rake db:migrate
+20. How does a model relate to a database?   
+A model is a table in the database
```
s-espinosa: A model represents a table in the database, but I wouldn't consider them the same thing. You can create and run a migration to create a table in the database, and it will exist, but you need to create the model separately so that you can use it in your application (if you plan on using ActiveRecord and not simply making direct calls to the DB).

module-2/backend/week_two.md
```diff
@@ -4,38 +4,89 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+ActiveRecord is a collection of methods written to streamline common Ruby methods. 
+
+2. What is a migration?  
+A migration is the instruction for how to alter/create tables in a database.   
+
+3. How does a table relate to a model?  
+A model represents a table in a database. You can build methods on a model to interact with a table.   
+
+4. What kind of methods are `belongs_to`, and `has_many`? (i.e. class or instance) Give an example.  
+They are class methods. Class and Teacher would have a one to many relationship, a class has one teacher and teachers have many classes. Students and Class would have a many to many relationships as a class has many students and students have many classes.
```
s-espinosa: I think we were trying to point out that you would need to have a specific instance of Class before you could use the methods that `belongs_to` offers you. For example:

```ruby
# Assuming our Class class has a `belongs_to :teacher` line
class_1704 = Class.find(1) # finds the specific class with an id of 1
mr_fritz      = class_1704.teacher # uses the teacher method that we now have to return a specific teacher associated with `class_1704`
```

module-2/backend/week_two.md
```diff
@@ -4,38 +4,89 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+They are class methods. Class and Teacher would have a one to many relationship, a class has one teacher and teachers have many classes. Students and Class would have a many to many relationships as a class has many students and students have many classes.   
+
+5. What do they allow you to do?  
+These associations allow tables to have relationships with other tables and the information within them to be accessible through their related tables.   
+
+6. What's the difference between agile workflow and waterfall method?  
+Agile workflow involves breaking apart a project into small chunks, completing all aspects of that chunk before moving on to another one. Waterfall moves through each stage of developmentin sequence for the whole project.  
+
+7. What is the difference between `#new` and `#create`?  
+New makes a new instance, create makes a new instance and saves it.
```

module-2/backend/week_two.md
```diff
@@ -4,38 +4,89 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+Agile workflow involves breaking apart a project into small chunks, completing all aspects of that chunk before moving on to another one. Waterfall moves through each stage of developmentin sequence for the whole project.  
+
+7. What is the difference between `#new` and `#create`?  
+New makes a new instance, create makes a new instance and saves it.   
+
+8. At a basic level, what does cURL allow you to do?  
+Pull in url from the command line.    
+
+9. In a database that's holding students and teachers, what will be the relationship between students and teachers? Draw the schema diagram.  
+Students and Teachers have a many to many relationship. Stu have many Tch and Tch have many stu.  They would be connected in a db through a join table i.e. students_teachers
```

module-2/backend/week_two.md
```diff
@@ -4,38 +4,89 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+Primary key is the key that is a unique identifier on a table.  
+Schema is the overall structure of the database - the tables, what is in them, and how they relate to each other.   
+
+11. Describe the relationship between a foreign key on one table and a primary key on another table.  
+A primary key become the foreign key on another table. For example if you had class and teacher (one to many), the class would have a foreign key which is comprised of teacher's primary key.   
+
+12. What are the parts of an HTTP response?  
+Status, Header, Body   
+
+13. `Rack::Test` allows us to test our controllers in isolation. What are some of the methods it gives us to simulate the request/response cycle?
```
s-espinosa: Think we removed this from the latest version. We didn't teach Rack::Test this time around, but you'll see controller testing in Mod 3.

module-2/backend/week_two.md
```diff
@@ -4,38 +4,89 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+12. What are the parts of an HTTP response?  
+Status, Header, Body   
+
+13. `Rack::Test` allows us to test our controllers in isolation. What are some of the methods it gives us to simulate the request/response cycle?  
+.to be_valid
+.to eq()
+.to not_eq
+
+14. Describe some techniques to make our Sinatra views more DRY. Give an example of when you would use these techniques.   
+You can make views more dry by extracting some of the view into a partial if it can be used in another view. For example, you can extract a form if it will be used in multiple spots.
```

module-2/backend/week_two.md
```diff
@@ -4,38 +4,89 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+8. What case does Ruby use for multi-word variables?  
+snake_case  
+
+9. In a database that's holding schools and teachers, what will be the relationship between schools and teachers?  
+One to many, a school has_many teachers and a teacher belongs_to a school.  
+
 10. In the same database, what will you need to do to create this relationship (draw a schema diagram)?
-11. Give an example of when you might want to store information besides ids on a join table.
+11. Give an example of when you might want to store information besides ids on a join table.  
+If there is information that is specific to each join combination.
```

PR #45, opened 2016-10-13
Update week_one.md

vidoseaver:
Sorry its so late

### DIFF
--- a/module-2/backend/week_two.md
+++ b/module-2/backend/week_two.md
@@ -5,17 +5,43 @@ Fork this respository. Answer the questions to the best of your ability. Try to 
 Note: When you're done, submit a PR. 
 
 1. At a high level, what is ActiveRecord? What does it do/allow you to do?
-2. What kind of methods are `belongs_to`, and `has_many`? (i.e. class or instance) Give an example.
-3. What do they allow you to do?
-4. What's the difference between agile workflow and waterfall method?
-5. What is the difference between `#new` and `#create`?
-6. At a basic level, what does cURL allow you to do?
-7. In a database that's holding students and teachers, what will be the relationship between students and teachers? Draw the schema diagram.
-8. Define foreign key, primary key, and schema.
-9. Describe the relationship between a foreign key on one table and a primary key on another table.
-10. What are the parts of an HTTP response?
-11. Describe some techniques to make our Sinatra code more DRY. Give an example of when you would use these techniques.
+Active record is a framework for SQL that has methods and allows data to be manipulated more easily.
 
+2. What kind of methods are `belongs_to`, and `has_many`? (i.e. class or instance) Give an example.
+pointers? I don't know/ forget the name -  but they point to other objects that they either belong to or have alot of.
+
+3. What do they allow you to do?
+They allow you to access the items from another table with the corresponding id.
+
+4. What's the difference between agile workflow and waterfall method?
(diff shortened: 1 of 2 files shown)


module-2/backend/week_two.md
```diff
@@ -5,17 +5,59 @@ Fork this respository. Answer the questions to the best of your ability. Try to
 Note: When you're done, submit a PR. 
 
 1. At a high level, what is ActiveRecord? What does it do/allow you to do?
+
+Active record is a framework for SQL that has methods and allows data to be manipulated more easily.
+
 2. What kind of methods are `belongs_to`, and `has_many`? (i.e. class or instance) Give an example.
+
+pointers? I don't know/ forget the name -  but they point to other objects that they either belong to or have alot of.
```
s-espinosa: I think the bigger question we were asking was what Casey went over in class today regarding what methods these relationships make available to you. Those methods will be instance methods. For example:

```ruby
blue_dog = Dog.find(1) # gives us a dog with the ID of 1.
red_house = blue_dog.house # Assigns the house associated with the `blue_dog` (a specific *instance* of Dog) to the variable red_house
```

module-2/backend/week_two.md
```diff
@@ -5,17 +5,59 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+
+They allow you to access the items from another table with the corresponding id.
+
 4. What's the difference between agile workflow and waterfall method?
+
+Agile is a closer production cycle that builds on itself. Waterfall is taking the long shot.
+
 5. What is the difference between `#new` and `#create`?
+
+new is the get method. create is  a post
```
s-espinosa: I think you're thinking of the methods that we use in our controllers. We were thinking of the ActiveRecord methods. When you call `#new` on an ActiveRecord model you create a new instance of that model, but it does not save to the database until you call `.save`. On the other hand, `#create` creates a new instance and saves it to the DB.

module-2/backend/week_two.md
```diff
@@ -5,17 +5,59 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+
 6. At a basic level, what does cURL allow you to do?
+
+Send stuff like it was being brought in from an outside ip? I think
```
s-espinosa: Yup. It does this by making HTTP requests to a server that you've booted up on your local machine. You can also make those requests from your machine to servers on the web.

module-2/backend/week_two.md
```diff
@@ -5,17 +5,59 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+
 9. Describe the relationship between a foreign key on one table and a primary key on another table.
+
+
+the primary key of teachers is the foriegn key in student
+
 10. What are the parts of an HTTP response?
+
+
+ you have the server, port then stuff and then params
```
s-espinosa: The response has headers and a body. The first header is the status code (e.g. 404, 200, 503, etc.).

module-2/backend/week_two.md
```diff
@@ -5,17 +5,59 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+ 
 11. Describe some techniques to make our Sinatra code more DRY. Give an example of when you would use these techniques.
 
+Make methods for stuff you're doing over and overagain in your controllers.  For tests build them in the test helper.
```

PR #53, opened 2016-10-18
Update week_two.md

bradgreen3:


module-2/backend/week_two.md
```diff
@@ -1,37 +1,56 @@
...
-Fork this respository. Answer the questions to the best of your ability. Try to answer them with limited amount of external research. These questions cover the majority of what we've learned this week (which is a TON - YOU are a web developer!!!). 
+Fork this respository. Answer the questions to the best of your ability. Try to answer them with limited amount of external research. These questions cover the majority of what we've learned this week (which is a TON - YOU are a web developer!!!).
 
-Note: When you're done, submit a PR. 
+Note: When you're done, submit a PR.
 
 1. At a high level, what is ActiveRecord? What does it do/allow you to do?
+It is a DSL that allows you to treat tables within a SQL database as Ruby objects.
 2. What kind of methods are `belongs_to`, and `has_many`? (i.e. class or instance) Give an example.
+Class methods. A horse belongs_to a stable, and a stable has_many horses.
```
case-eee: i think this question is misleading - i'm more interested in how you use the methods these associations give you (we talked about this in class today). in your example with `horses` and `stables`. the methods that these associations give you are instance methods 👍

module-2/backend/week_two.md
```diff
@@ -1,37 +1,56 @@
...
+Create associations between two tables.
 4. What's the difference between agile workflow and waterfall method?
+Waterfall is sequential--planning, designing, creating, testing. You know the outcome you want beforehand. Agile is incremental. You tackle a project in iterations, for instance. You don't typically know what the outcome until you get there.
 5. What is the difference between `#new` and `#create`?
+These are methods. #new usually takes you to a form to create a new thing, whereas #create is the method called when pressing a submit button on said form.
 6. At a basic level, what does cURL allow you to do?
+Send/receive information to/from a server.
 7. In a database that's holding students and teachers, what will be the relationship between students and teachers? Draw the schema diagram.
+I understand that a join table is needed for this many-to-many relationship, but I don't fully understand why. Nevertheless, students has_many students_teachers and has_many :teachers, through: :students_teachers. Students_teachers belongs_to :students and belongs_to :teachers. Teachers has_many students_teachers and has_many students, through: :students_teachers.
```
case-eee: it's necessary to track the relationships between the two tables. if we didn't have it, we wouldn't be able to track the `many-to-many` piece

module-2/backend/week_two.md
```diff
@@ -1,37 +1,56 @@
...
 8. Define foreign key, primary key, and schema.
+A foreign key is an ID linking an object within one table to an object within another. A primary key is assigned to each new object/row that is created within a table. In short, it references an object within a single, local table. A schema is a file that displays the tables and their attributes/columns within a database.
 9. Describe the relationship between a foreign key on one table and a primary key on another table.
+Each object within a table will(?) have a primary key. If those objects are linked to an object within another table, then they will be linked via the primary key of the foreign object. This is a foreign key. That was more difficult to explain than it should've been.
```
case-eee: i think your explanation is fine. it's hard to explain with words. the primary key on one table will directly relate to the foreign key on another table. they hold the same value 👍

PR #55, opened 2016-10-23
Update and rename week_three.md to Anthony_week_three.md

antciccone:


module-2/backend/Anthony_week_three.md
```diff
@@ -0,0 +1,70 @@
...
+ * resources :horses, only: [:show]
+
+5. What rake task is useful when looking at routes, and what information does it give you?
+ * rake routes - it gives you info to what paths are routes are available  
+
+6. What is an example of a route helper? When would you use them?
+ * a route helper is really useful when you are trying to redirect or link to a new page. For example if you want to link to the index for Companies it would be companies_path  
+
+7. What's the difference between what `_url` and `_path` return when combined with a routes prefix?
+  * hmm not totally sure. In all the projects so far I have only linked to or redirected to something that ends with _path not _url
```
s-espinosa: `_url` gives us the absolute address (basically the full URL including `http://domainname.com/`). `_path` only gives us the relative path.

module-2/backend/Anthony_week_three.md
```diff
@@ -0,0 +1,70 @@
...
+  * hmm not totally sure. In all the projects so far I have only linked to or redirected to something that ends with _path not _url  
+
+8. What are strong params and why are the necessary?
+ * Strong params usually go in ur create and to add a level of security so that the user can not mess with the database.
+
+9. What role does `form_for` play in helping us create our forms?
+ * Rails helper method that created a form for your user to input data
+
+10. How does `form_for` know where to submit the user's input?
+  *.label is what you title that input and .text_field :(what ever column you want this info to be assigned to)
```
s-espinosa: True, but I think we were thinking more along the lines of what path the form submits to (take a look at the source for a form created by one of your Rails apps if this is confusing). `form_for` uses the instance that you pass to it as an argument to make this determination. So, for example, if you pass it @horse, and @horse is an instance of your `Horse` class that has not yet been saved to the database, `form_for` will create a form that submits to `/horses` with a POST. If we look at our routes, this means that the `#create` method in our `HorsesController` will handle the request.

module-2/backend/Anthony_week_three.md
```diff
@@ -0,0 +1,70 @@
...
+    ```
+    In new.html.erb
+    ```
+      <h1> Create a New Horse<h1>
+      <%= form_for @horse do |f| %>
+      <%= f.label "Name" %>
+      <%= f.text_field :name%>
+      <%= f.submit %>
+      <%end%>
+    ```
```
s-espinosa: I like that you include the controller piece of this! :tada:

module-2/backend/Anthony_week_three.md
```diff
@@ -0,0 +1,70 @@
...
+ * They generally inherit from ActiveRecord::Base
+
+3. What do Controllers generally inherit from in a rails project?
+ * They generally inherit from ApplicationController
+
+4. How would I create a route if I wanted to see a specific horse in my routes fitle assuming I'm sticking to standard conventions and that I didn't want other CRUD functionality?
+ * resources :horses, only: [:show]
+
+5. What rake task is useful when looking at routes, and what information does it give you?
+ * rake routes - it gives you info to what paths are routes are available
```
s-espinosa: Yup! To add a little more detail it shows you prefixes, verbs, URI patterns, and controller actions.

PR #57, opened 2016-10-24
Week 3

epintozzi:


module-2/backend/week_three.md
```diff
@@ -10,14 +10,83 @@ Note: When you're done, submit a PR with a reflection in the comments about how
 ### Questions
 
 1. What is the entry at the command line to create a new rails app?
+<br>
+`rails new project_name`
```

module-2/backend/week_three.md
```diff
@@ -10,14 +10,83 @@ Note: When you're done, submit a PR with a reflection in the comments about how
...
+<br>
+<br>
+
 4. How would I create a route if I wanted to see a specific horse in my routes fitle assuming I'm sticking to standard conventions and that I didn't want other CRUD functionality?
+<br>
+If you only wanted the `index` route for example:
+`resources :horse, only: [:index]`
+or
+`get '/horse' => 'horse#index'`
+<br>
```
s-espinosa: Like that you included both options! 🎉

module-2/backend/week_three.md
```diff
@@ -10,14 +10,83 @@ Note: When you're done, submit a PR with a reflection in the comments about how
...
+<br>
+
 5. What rake task is useful when looking at routes, and what information does it give you?
+<br>
+`rake routes` shows you all the routes you have available, as well as the prefix, verb, url patter, and which controller and method you'd need to use for each.
```

module-2/backend/week_three.md
```diff
@@ -10,14 +10,83 @@ Note: When you're done, submit a PR with a reflection in the comments about how
...
+
 9. What role does `form_for` play in helping us create our forms?
+<br>
+`form_for` is a shortcut that helps us create forms in rails. 
+<br>
+<br>
+
 10. How does `form_for` know where to submit the user's input?
+<br>
+It knows depending on which http verb we're using.
```
s-espinosa: Interesting! It's true that rails knows what to do with input from a form based on the verb/URI combination, but I think we were one step back from that: how does `form_for` know what to put in the `action` field (and the hidden `method` field for an update) when it generates the form in HTML?

It uses the argument passed to it (e.g. `@horse` in the example you provide below), and uses it to generate a path and determine the proper method to hit the right route to either create or update. Let us know if this makes sense or if you'd like to discuss! 🎉

module-2/backend/week_three.md
```diff
@@ -10,14 +10,83 @@ Note: When you're done, submit a PR with a reflection in the comments about how
...
+  <%= f.submit %>
+<% end %>
+
+```
+<br>
+<br>
+
 12. Why do we want to validate our models?
+<br>
+We validate our models to make sure they contain all the required attributes when they are created and to ensure uniqueness (when necessary).
```

PR #58, opened 2016-10-24
Update week_three.md

AliSchlereth:
I felt pretty comfortable with the questions this week. While I did not look back at my notes, I was sorely tempted to to make sure that my answers were correct. I hope to get more confident as we continue the mod.

### DIFF
--- a/module-2/backend/week_three.md
+++ b/module-2/backend/week_three.md
@@ -10,14 +10,45 @@ Note: When you're done, submit a PR with a reflection in the comments about how 
 ### Questions
 
 1. What is the entry at the command line to create a new rails app?
-2. What do Models generally inherit from in rails?
+rails new  
+
+2. What do Models generally inherit from in rails?  
+ActiveRecord::Base
+
 3. What do Controllers generally inherit from in a rails project?
-4. How would I create a route if I wanted to see a specific horse in my routes fitle assuming I'm sticking to standard conventions and that I didn't want other CRUD functionality?
-5. What rake task is useful when looking at routes, and what information does it give you?
-6. What is an example of a route helper? When would you use them?
-7. What's the difference between what `_url` and `_path` return when combined with a routes prefix?
-8. What are strong params and why are the necessary?
-9. What role does `form_for` play in helping us create our forms?
-10. How does `form_for` know where to submit the user's input?
-11. Create a form using a `form_for` helper to create a new `Horse`. 
-12. Why do we want to validate our models?
+ApplicationController  
+
+4. How would I create a route if I wanted to see a specific horse in my routes fitle assuming I'm sticking to standard conventions and that I didn't want other CRUD functionality?  
+resources :horses, only [:show]  
+
+5. What rake task is useful when looking at routes, and what information does it give you?  
+rake routes  
+
+6. What is an example of a route helper? When would you use them?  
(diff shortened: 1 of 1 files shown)


module-2/backend/week_three.md
```diff
@@ -10,14 +10,45 @@ Note: When you're done, submit a PR with a reflection in the comments about how
...
-8. What are strong params and why are the necessary?
-9. What role does `form_for` play in helping us create our forms?
-10. How does `form_for` know where to submit the user's input?
-11. Create a form using a `form_for` helper to create a new `Horse`. 
-12. Why do we want to validate our models?
+ApplicationController  
+
+4. How would I create a route if I wanted to see a specific horse in my routes fitle assuming I'm sticking to standard conventions and that I didn't want other CRUD functionality?  
+resources :horses, only [:show]  
+
```

module-2/backend/week_three.md
```diff
@@ -10,14 +10,45 @@ Note: When you're done, submit a PR with a reflection in the comments about how
...
+5. What rake task is useful when looking at routes, and what information does it give you?  
+rake routes
```
s-espinosa: Yup! With regard to what does it give you:

* Prefix - building block for your url/path helpers
* Verb - HTTP verb to use when accessing this route
* URI Pattern - Path that a user would enter in their nav bar to access the resource including any wildcards
* Controller#Action - The controller and method on that controller that will handle this request.

module-2/backend/week_three.md
```diff
@@ -10,14 +10,45 @@ Note: When you're done, submit a PR with a reflection in the comments about how
...
+_path gives you the relative path while _url gives you the whole url  
+
+8. What are strong params and why are the necessary?  
+This checks to make sure that objects are getting created with the correctly correlated info.  
+
+9. What role does `form_for` play in helping us create our forms?  
+It builds the html form as well as passes the information on to create/update tied to the right keys.  
+
+10. How does `form_for` know where to submit the user's input?  
+..It uses the input instance variable in combination with the type of page it came from (new/edit)
```
s-espinosa: For 9 & 10: `form_for` doesn't actually pass the information, it just creates the HTML form that will be rendered on the user's machine along with all the necessary pieces to send it to the right place.

It knows where to send the users input based on the _instance_ that we pass and whether or not that has been saved in our database (`User.new` has not been saved to the database, but `User.find(1)` has). [This](http://stackoverflow.com/questions/11868489/how-does-rails-decide-to-render-a-form-with-a-method-of-put-or-post) StackOverflow answer has a decent description. I think it's also helpful to inspect a page with a form that Rails has created to help you get a sense of how this all comes together! 👍

module-2/backend/week_three.md
```diff
@@ -10,14 +10,45 @@ Note: When you're done, submit a PR with a reflection in the comments about how
...
+11. Create a form using a `form_for` helper to create a new `Horse`.   
+form_for @horse do |f|  
+f.label :name  
+f.text_field :name  
+
+f.label :age
+f.text_field :age  
+
+f.submit
+end
```
s-espinosa: So interesting that you didn't include the erb tags! I think it highlights that this is just a Ruby method, which is super nice to know. I think you probably know this, but remember that the first line includes an `=` in the erb, while the closing `end` does not. It's one piece that sometimes gets a little tricky.

PR #60, opened 2016-10-24
Update week_three.md

rtravitz:


module-2/backend/week_three.md
```diff
@@ -10,14 +10,63 @@ Note: When you're done, submit a PR with a reflection in the comments about how
...
+
+
 6. What is an example of a route helper? When would you use them?
+
+A route helper is something like horses_path, which would bring you to the index of horses. You use them in place of writing out a url.
+
+
 7. What's the difference between what `_url` and `_path` return when combined with a routes prefix?
+
+path is relative. url is absolute.
```

module-2/backend/week_three.md
```diff
@@ -10,14 +10,63 @@ Note: When you're done, submit a PR with a reflection in the comments about how
 ### Questions
 
 1. What is the entry at the command line to create a new rails app?
+
+rails new (And maybe the option to not use Spring. Spring breaks dreams.)
```
s-espinosa: So many broken dreams. 😿

module-2/backend/week_three.md
```diff
@@ -10,14 +10,63 @@ Note: When you're done, submit a PR with a reflection in the comments about how
...
+
+
 9. What role does `form_for` play in helping us create our forms?
+
+form_for makes it easy to create forms to create and update ActiveRecord objects.
+
 10. How does `form_for` know where to submit the user's input?
-11. Create a form using a `form_for` helper to create a new `Horse`. 
+
+An instance of a model is passed in when form_for is called. It creates or updates that instance.
```
s-espinosa: Right general idea, but it's important to think about how this whole cycle works. `form_for` will take the parameter that's passed to it and generate an HTML form that will submit to the appropriate route using the appropriate method (HTTP verb). Your URI pattern in your `rake routes` output will tell you what that route/method is. View source on one of the forms you create to see this in action.

module-2/backend/week_three.md
```diff
@@ -10,14 +10,63 @@ Note: When you're done, submit a PR with a reflection in the comments about how
...
+  <%= f.label :name %>
+  <%= f.text_field :name %>
+  
+  <%= f.submit %>
+<% end %>
+```
+
 12. Why do we want to validate our models?
+
+We want to validate our models to make sure that critical information is not left blank. In my experience so far, this can cause problematic nil errors in views.
```

PR #61, opened 2016-10-24
Anna Dolan week three checks for understanding

annadolan:
I felt on top of most of these questions, although things like "what are strong params for" kind if threw me. I feel as though our project this weekend really cemented some of these ideas in my mind.


module-2/backend/week_three.md
```diff
@@ -1,25 +1,64 @@
...
+  
+    A: ActiveRecord::Base
+  
+3. What do Controllers generally inherit from in a rails project?
+
+    A: ApplicationController
+    
+4. How would I create a route if I wanted to see a specific horse in my routes fitle assuming I'm sticking to standard conventions and that I didn't want other CRUD functionality?
+
+    A: resources :horses, only: [:show]
```
s-espinosa: 🎉

module-2/backend/week_three.md
```diff
@@ -1,25 +1,64 @@
...
+    A: For the route in question 4, you would use horse_path(horse) as a route helper. They are helpful when you want to  
+      redirect or link within your app and are using RESTful conventions.
+      
+7. What's the difference between what `_url` and `_path` return when combined with a routes prefix?
+
+    A: `_url` returns the full URL, `_path` only returns the relative path.
+    
+8. What are strong params and why are the necessary?
+
+    A: Strong params allow you to choose params that can be used in controller methods.
```
s-espinosa: This is true, inasmuch as we're using them in our `create` and `new` methods, but the key here is that they specify which attributes we're willing to accept from users in our forms. For example, if a user is creating a new login and we store `role` in our database, we might not want to let them submit that as part of their form, so we wouldn't include it in our strong params. It's a safeguard against people editing the HTML of our site in their browser and submitting things that we don't want them to be able to submit. 👍

module-2/backend/week_three.md
```diff
@@ -1,25 +1,64 @@
...
+    
+9. What role does `form_for` play in helping us create our forms?
+    
+    A: It makes it easier to make forms in general, and it also allows us to use the same partial for "new" and "edit".
+    
+10. How does `form_for` know where to submit the user's input?
```
s-espinosa: `form_for` uses the argument that we passed to it (often an instance variable holding either a new or existing instance of an object) to determine the route and verb that will be used in the HTML of the form it generates. I think sometimes this feels a little bit like magic, so next time you get a chance inspect the raw HTML of a form in one of your rails apps. 👍

module-2/backend/week_three.md
```diff
@@ -1,25 +1,64 @@
...
+11. Create a form using a `form_for` helper to create a new `Horse`. 
+    ```erb
+    <% @horse = Horse.new(horse_params) %>
+    <%= form_for @horse do |f| %>
+    <%= f.label :name %>
+    <%= f.text_field :name %>
+    
+    <%= f.submit %>
+    <% end %>
+    ```
```
s-espinosa: 🎉

PR #63, opened 2016-12-02
Update week 1

s-espinosa:
@case-eee Just a couple of extra questions that were previously in the week 2 set. Let me know what you think!


case-eee:
@s-espinosa looks good to me. i'm going to act like i don't see what time this PR came in at

PR #64, opened 2016-12-02
Update week two

s-espinosa:
@case-eee Submitting as a separate request to keep the week 1 and week 2 changes in their own PRs.

* Removes questions added to week 1.
* Removes the cURL question.
* Proposes a replacement for the `belongs_to`/`has_many` question.

Let me know what you think!

### DIFF
--- a/module-2/backend/week_two.md
+++ b/module-2/backend/week_two.md
@@ -5,11 +5,28 @@ Fork this respository. Answer the questions to the best of your ability. Try to 
 Note: When you're done, submit a PR. 
 
 1. At a high level, what is ActiveRecord? What does it do/allow you to do?
-2. What kind of methods are `belongs_to`, and `has_many`? (i.e. class or instance) Give an example.
+2. Assume you have the following model:
+
+```ruby
+class Team << ActiveRecord::Base
+end
+```
+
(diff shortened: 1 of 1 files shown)


module-2/backend/week_two.md
```diff
@@ -5,11 +5,28 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
 1. At a high level, what is ActiveRecord? What does it do/allow you to do?
-2. What kind of methods are `belongs_to`, and `has_many`? (i.e. class or instance) Give an example.
+2. Assume you have the following model:
+
+```ruby
+class Team << ActiveRecord::Base
+end
+```
+
+What are some methods can you call on `Team`? If these methods aren't defined in the class, how do you have access to them?
```
case-eee: "What are some methods you can call on `Team`?"

module-2/backend/week_two.md
```diff
@@ -5,11 +5,28 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+
+4. Assume that you added a line to your `Team` class as follows:
+
+```ruby
+class Team << ActiveRecord::Base
+  belongs_to :owner
+end
+```
+
+Now how would you find the owner of the team with an id of 4?
```
case-eee: 👍

PR #68, opened 2016-12-05
Update week_one.md

cews7:
Great first week!


module-2/backend/week_one.md
```diff
@@ -5,30 +5,105 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
 2. What is Sinatra?
+
+* Sinatra is a web-application framework. It allows the developer to configure things to his liking more than Rails.
 4. What is MVC?
+
+* Model, View, Controller. Model: holds our logic; Controller: serves as our router; View: displays content
+
 5. Why do we follow conventions when creating our actions/path names in our Sinatra routes?
+
+* To be in sync with what Rails does?
```
s-espinosa: To some degree this is definitely true. We're using Sinatra as a stepping stone. I think that we were thinking about this in terms of conventions in a more general sense, which allow us to step into an unfamiliar code base and quickly get up to speed so we can contribute (and allows others to do the same in our code base).

module-2/backend/week_one.md
```diff
@@ -5,30 +5,105 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+
 7. Given the following block of code, how would I pass an instance variable `count` with a value of `1` to my `index.erb` template?
   
   ```ruby
   get '/horses' do
     erb :index
   end
+  
+  get '/names' do
+    erb :index, :locals => {:name => params[:Mr.Ed]}
```
s-espinosa: Need to include `@count = 1` in this block to pass the instance variable so that we can use it in the erb file.

module-2/backend/week_one.md
```diff
@@ -5,30 +5,105 @@ Fork this respository. Answer the questions to the best of your ability. Try to
 Note: When you're done, submit a PR. 
 
 1. List the five common HTTP verbs and what the purpose is of each verb.
+
+* DELETE: delete , GET: render a view, POST: submit content, PUT: update content, PATCH: sililar to PUT, used in Rails.
```

module-2/backend/week_one.md
```diff
@@ -5,30 +5,105 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
 
 1. List the five common HTTP verbs and what the purpose is of each verb.
+
+* DELETE: delete , GET: render a view, POST: submit content, PUT: update content, PATCH: sililar to PUT, used in Rails.
 2. What is Sinatra?
+
+* Sinatra is a web-application framework. It allows the developer to configure things to his liking more than Rails.
 4. What is MVC?
+
+* Model, View, Controller. Model: holds our logic; Controller: serves as our router; View: displays content
```

module-2/backend/week_one.md
```diff
@@ -5,30 +5,105 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+  end
+  
+  index.erb
+  <% @count %>
   ```
 
 8. In the same code block, how would I pass a local variable `name` with a value of `Mr. Ed`?
 9. What's the purpose of ERB?
+
+* Make our Ruby code readible by the browser.
```
s-espinosa: Small note: our ERB files get parsed server side, so that by the time we send them to the browser all it's seeing is HTML. I would say that the above is correct, but wanted to make sure we had the full picture.

module-2/backend/week_one.md
```diff
@@ -5,30 +5,105 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+
 10. Why do I need a development AND test database?
+
+* You don't want your production database having content from your test database. In other words, you wouldn't want the samples you create in testing to end up in production.
```
s-espinosa: In production, for sure, but also even in our *development* database that we run locally where sometimes we'll want to keep some information, but not mix that with our tests.

module-2/backend/week_one.md
```diff
@@ -5,30 +5,105 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+
+* put erb :index 
+
+* get erb :show 
+
+* delete erb :index
+
+* post erb :show 
+
+* get erb :edit
```
s-espinosa: Regarding the routes: we were looking for the URL patterns, so for example a `GET` request to `/restaurants` should show us a list of all restaurants. A `POST` to `/restaurants` will create a new restaurant, etc.

module-2/backend/week_one.md
```diff
@@ -5,30 +5,105 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+
+* get erb :edit 
+
 18. What's a migration? 
+
+* creating a database schema or adding to an existing one
+
 19. When you create a migration, does it automatically modify your database?
+
+* yes?
```
s-espinosa: Nope! We have to run `rake db:migrate`.

module-2/backend/week_one.md
```diff
@@ -5,30 +5,105 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+
 21. What's the difference between agile workflow and waterfall method?
+
+* agile: go through entire process of planning and executing a piece of functionality, whereas waterfall entails building the entire product at once with no feedback from users until a full product is realized.
+
 22. What is the difference between `#new` and `#create`?
+
+* 'new makes an instance, but doesn't save it. 
+
+* 'create' makes and instance and saves it in the database.
```

module-2/backend/week_one.md
```diff
@@ -5,30 +5,105 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+
+* creating a database schema or adding to an existing one
+
 19. When you create a migration, does it automatically modify your database?
+
+* yes? 
+
 20. How does a model relate to a database?
+
+* through Active Record. The devloper can set different relationships from within the model
```
s-espinosa: Small clarification: I would say that the relationships are set at the database level using foreign keys. The models provide us with methods that we can use to take advantage of those relationships at the model/object level in our app.

PR #95, opened 2016-12-12
Nick Erhardt week 2 check for understanding

ski-climb:


module-2/backend/week_two.md
```diff
@@ -5,6 +5,8 @@ Fork this respository. Answer the questions to the best of your ability. Try to
 Note: When you're done, submit a PR. 
 
 1. At a high level, what is ActiveRecord? What does it do/allow you to do?
+ * *ActiveRecord is the go-between from ruby to the database.  It allows us to query the database using ruby and ruby-like (ActiveRecord) methods.*
+ * *It also allows us to swap out what kind of database (e.g. Postgresql, MYSQL, etc...) we're using and not change our model or migration code, although I'm not sure what kind of changes would have to be made if we swapped to a non-relational database.*
```

module-2/backend/week_two.md
```diff
@@ -13,8 +15,13 @@ end
 ```
 
+ * *`Team.all, Team.find, .find_by, .count, .where, .pluck, .find_each, ...`*
+ * *We have access to those methods via inheriting from ActiveRecord::Base - which looks into the database and does something called 'reflection' to provide us with methods based on what columns that table has in the database, AND provides some standard methods liks those listed above.*
```

module-2/backend/week_two.md
```diff
@@ -25,13 +32,33 @@ end
 ```
 
+ * *`owner_name = Team.find(4).owner.name`*
```

module-2/backend/week_two.md
```diff
@@ -25,13 +32,33 @@ end
...
+class Teacher << ActiveRecord::Base
+  has_many :students
+end
+```
+
+ * *As far as the schema, the Student table would have an id, a foreign key called teacher_id, and most likely a name and maybe a school assignd student id.  The Teacher table would have an id and most likely a name.*
+
 8. Define foreign key, primary key, and schema.
+ * *A foreign key is a column on a database table that points to another row's primary key (`teacher_id` is a foreign key in the above example).*
+ * *A primary key is the unique identifier for a row within a database table.  It is often an integer, but only needs to be a column o unique values (think Social Security Numbers here).*
```

PR #148, opened 2017-01-28
Update week_one.md

valerietrudell:


module-2/backend/week_one.md
```diff
@@ -5,30 +5,99 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
-6. What types of variables are accessible in our view templates without explicitly passing them?
+    a. Model
+    b. View 
+    c. Controller 
+    This is the file structure used in sinatra/ruby programs
+5. Why do we follow conventions when creating our actions/path names in our Sinatra routes? 
+      Because it allows for easy understanding across all programs. It is saved time that would otherwise be spend figuring  
+      out what extension goes to what page. 
+6. What types of variables are accessible in our view templates without explicitly passing them? 
+      yield
```
s-espinosa: `yield` isn't really a variable here. It's acting as a keyword. We were looking for *instance variables* (consistent with your answer to question 7).

module-2/backend/week_one.md
```diff
@@ -5,30 +5,99 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
-  get '/horses' do
+  get '/horses' do 
     erb :index
   end
   ```
-
-8. In the same code block, how would I pass a local variable `name` with a value of `Mr. Ed` to the view?
-9. What's the purpose of ERB?
-10. Why do I need a development AND test database?
+    Add => @count = 1 above erb:index
```

module-2/backend/week_one.md
```diff
@@ -5,30 +5,99 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
   end
   ```
-
-8. In the same code block, how would I pass a local variable `name` with a value of `Mr. Ed` to the view?
-9. What's the purpose of ERB?
-10. Why do I need a development AND test database?
+    Add => @count = 1 above erb:index 
+    
+8. In the same code block, how would I pass a local variable `name` with a value of `Mr. Ed` to the view? 
+    Add => name = "Mr. Ed"
```
s-espinosa: Part way there! But we also need to change the `erb :index` line to the following:

```ruby
get '/horses' do
  name = "Mr. Ed"
end
```

Alternatively we could also do something like the following, which skips the step of declaring the local variable in the `get` method:

```ruby
get '/horses' do 
  erb :index, locals: {name: "Mr. Ed"}
end
```

module-2/backend/week_one.md
```diff
@@ -5,30 +5,99 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+    To allow ruby and html/css to interact together 
+10. Why do I need a development AND test database? 
+    Development is just set/moving things in and around 
+    Test is creating instances that we don't want to stick around after we have tested a thing. 
 11. What's responsive design?
-12. What is CRUD and why is it important?
-13. What does HTTP stand for? 
-14. What are the two ways to interpolate Ruby in an ERB view template? What's the difference between these two ways?
-15. What's an ORM?
+    When you resize a browser window and the site respondse in a way that still makes sense to the user viewing the new size
```

module-2/backend/week_one.md
```diff
@@ -5,30 +5,99 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+    
+    These are important because it explains all of the functionality of a website. There isn't anything more that needs to be 
+    done besides C/R/U/D
+13. What does HTTP stand for?  
+    Hyper
+    Text 
+    Transfer
+    Protocol
+14. What are the two ways to interpolate Ruby in an ERB view template? What's the difference between these two ways? 
+    adding <%= @name %> or #{@name}
```
s-espinosa: First part is right, but the second way is to have `<% %>` without the equals sign. The difference is that with the equals sign the code inside will display on the page, and without it will not.

module-2/backend/week_one.md
```diff
@@ -5,30 +5,99 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+15. What's an ORM? 
+    Object 
+    Relational
+    Mapping
```
s-espinosa: Yup! But think the question was asking more what they do. Might already have this down, but just to make sure, an ORM is a tool that helps us interact with our databases.

module-2/backend/week_one.md
```diff
@@ -5,30 +5,99 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
 16. What's the most commonly used ORM in ruby (Sinatra & Rails)?
-17. Let's say we have an application with restaurants. There are seven verb + path combinations necessary to provide full CRUD functionality for our restaurant application. List each of the seven combinations, and explain what each is for.
-18. What's a migration? 
-19. When you create a migration, does it automatically modify your database?
+    Rails
```
s-espinosa: ActiveRecord.

module-2/backend/week_one.md
```diff
@@ -5,30 +5,99 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
-19. When you create a migration, does it automatically modify your database?
+    Rails 
+17. Let's say we have an application with restaurants. There are seven verb + path combinations necessary to provide full CRUD functionality for our restaurant application. List each of the seven combinations, and explain what each is for. 
+    1. /restaurants  
+        Get 
+        - gets :index and displays
+    2.  /restaurants/new  
+        Get 
+        - gets :new and displays, user can enter in form
+    3. /restaurants/new
```
s-espinosa: The route for this is actually `/restaurants`.

We only use the `/new` to get the form to create a new restaurant.

module-2/backend/week_one.md
```diff
@@ -5,30 +5,99 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+        - gets :new and displays, user can enter in form
+    3. /restaurants/new  
+        Post 
+        - once user has selected Submit, the :new form is read and posted to :index 
+    4. /restaurants/:id  
+        Get 
+        - gets :show and displays selected resturant
+    5. /restaurants/:id/edit 
+        Get  
+        - gets :edit and displays the edit form to user where they can enter in or change information
```

module-2/backend/week_one.md
```diff
@@ -5,30 +5,99 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+    6. /restaurants/:id/edit
```
s-espinosa: Similar concept to `/new`. We send the POST request to `/restaurants/:id`.

module-2/backend/week_one.md
```diff
@@ -5,30 +5,99 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+    4. /restaurants/:id  
+        Get 
+        - gets :show and displays selected resturant
+    5. /restaurants/:id/edit 
+        Get  
+        - gets :edit and displays the edit form to user where they can enter in or change information
+    6. /restaurants/:id/edit 
+        Post  
+        - once Submmit is selected, form is read and posted to :index
+    7. /restaurants/:id
```

module-2/backend/week_one.md
```diff
@@ -5,30 +5,99 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+        Delete 
+        - takes the selected restuarant out of the list of created items and redirects to :index
+    
+    
+18. What's a migration?  
+      Any change or update made in a SQL database
```
s-espinosa: I would think of the migrations as the **instructions** to create/update a database.

module-2/backend/week_one.md
```diff
@@ -5,30 +5,99 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+19. When you create a migration, does it automatically modify your database?  
+      No you have to go into migration folder and put in the change or update desired
```
s-espinosa: Yup! Even after that, you'll need to run the migration with `rake db:migrate`.

module-2/backend/week_one.md
```diff
@@ -5,30 +5,99 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+      The models are the objects we are working with
+21. What's the difference between agile workflow and waterfall method? 
+      1. Waterfall: creates a plan from the begginning and sticks to that plan. A product is made and testing is completed,    
+      then after the product is reviewed.  
+      
+      2. Agile workflow: is taking the project, breaking into portions. You complete the most basic functions at first, the review and make changes. Then add an iteration that adds functionality which is then reviewed. Each iteration added makes a further path toward what was desired to begin with. This way allows the devs and clients to be on the same understanding, clients are mpre involved with the decision making process, and the final product is the thing the truely imagined to begin with. 
+22. What is the difference between `#new` and `#create`? 
+
+      1. #new just make a new instance of a thing. 
+      2. #create makes a new instance and also saves it
```

PR #150, opened 2017-01-28
Update week_one.md

brandonrandall:


module-2/backend/week_one.md
```diff
@@ -5,30 +5,112 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
     erb :index
   end
   ```
 
+
+  ```ruby
+  get '/horses' do
+  @count = 1
+  @name = Mr. Ed
```
s-espinosa: this would get the information to the view, but the question is asking how we would pass a **local variable**. When you put the `@` in front of `name`, you're declaring an **instance** variable. To pass a local variable, we would use the following syntax:

```ruby
get '/horses' do
  name = "Mr. Ed"
end
```

module-2/backend/week_one.md
```diff
@@ -5,30 +5,112 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+
+  object relational mappers: converts data between incompatible type systems in object oriented programming languages
+
 16. What's the most commonly used ORM in ruby (Sinatra & Rails)?
+
+  ActiveRecord
+
 17. Let's say we have an application with restaurants. There are seven verb + path combinations necessary to provide full CRUD functionality for our restaurant application. List each of the seven combinations, and explain what each is for.
+
+get '/'
```
s-espinosa: Make sure to read the questions carefully. We asked specifically about restaurants, so I would have that in place of `whatever` in your answers below.

module-2/backend/week_one.md
```diff
@@ -5,30 +5,112 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+get '/whatever'
+get '/whatever/new'
+get '/whatever/:id'
+get '/whatever/:id/edit'
+put '/whatever/:id'
+delete '/whatever/:id'
```
s-espinosa: Looks like we're missing a POST to '/restaurants' (GET '/' isn't part of the CRUD functionality). Also, you need to describe what each of these does!

module-2/backend/week_one.md
```diff
@@ -5,30 +5,112 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+
 18. What's a migration? 
+
+  a migration is an update of the table; creating changes to table based on info from the database
+
 19. When you create a migration, does it automatically modify your database?
+
+  yes(or does it modify the table?)
```
s-espinosa: It doesn't **automatically** modify the table. You need to run `rake db:migrate`.

module-2/backend/week_one.md
```diff
@@ -5,30 +5,112 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+  the model creates methods that will access the database
+
 21. What's the difference between agile workflow and waterfall method?
+
+  waterfall: plan at the start of the project and no deviation/variation from that
+  agile: continuous reiteration during a project
+
 22. What is the difference between `#new` and `#create`?
+
+  new does not save the new instance whereas create will create and save at the same time.
```

module-2/backend/week_one.md
```diff
@@ -5,30 +5,112 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+
+  Models, Views, Controllers
+
+5. Why do we follow conventions when creating our actions/path names in our Sinatra routes? 
+
+  they accomplish a simplified way to read the path names and make connections between desired actions
+
 6. What types of variables are accessible in our view templates without explicitly passing them?
+
+  the variables can represent the methods in the models, routes
```
s-espinosa: We wouldn't normally store a method in a variable, but we could save an instance of the model itself, which would then give us access to its methods.

Re-read the question, though. This is asking what **type** of variables we can access in our views. The answer to that is **instance variables** (which you show in your answer to number 7).

brandonrandall:
i didnt see a update pull request so im suspicious i didnt do the pull request correctly

PR #152, opened 2017-01-29
finishes courtney meyerhofer

meyerhoferc:
Mod 2 Week 1 Check for Understanding


module-2/backend/week_one.md
```diff
@@ -5,30 +5,123 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+     * GET '/restaurants/:id' 
+ 
+ Update
+   * User can update an existing restaurant
+     * GET '/restaurants/:id/edit' (retrieves form)
+     * PUT '/restaurants/:id' (updates to db, displays new data on show view)
+ 
+ Delete
+   * User can delete a restaurant
+     * DELETE '/restaurants/:id' (deletes from db, displays index view)
```
s-espinosa: This looks good! If you want more practice with this (per your response to question 1), I would recommend sketching this out with different resources (books, cats, teachers, schools, horses, muffins, cookies, etc.) when you have a break.

One small note: the POST, PUT, and DELETE actions don't actually display a new view; they redirect to other actions that do. For example, your POST '/restaurants' route adds a new restaurant, but then in order to display an index it would actually send instructions to the browser to make another request for a GET '/restaurants'. That's fairly nuanced, but something to be aware of.

Also, the particular view that is rendered after a PUT/POST/DELETE isn't necessarily part of this rule. You've given examples of what would be good practice, but if we wanted to send a user to a restaurant show page after they create a new restaurant that wouldn't be bad either.

module-2/backend/week_one.md
```diff
@@ -5,30 +5,123 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+
 18. What's a migration? 
+
+  a process that adds structure to an existing database
+ 
 19. When you create a migration, does it automatically modify your database?
+
+ not unless ``` rake db:migrate ``` has been executed
```

module-2/backend/week_one.md
```diff
@@ -5,30 +5,123 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+
 20. How does a model relate to a database?
+
+ A model is a representation of aspects of the database and tables within the database (or perhaps the other way around?). A model is a ruby object that, because it can inherit from ActiveRecrd can be represented by a database. Any attributes we would want this model to have as a Ruby object translate to columns of the table, and instantiations of the model's class are represented by rows of the table. The model also has behavior in that it can interact with other models via foreign keys in the database.
```
s-espinosa: I like the first way you describe this better than the second. To me, the database is the source of truth for our application, and the models are ruby objects that we use to interact with the database.

One other note: the interaction with other models isn't automatic once we have foreign keys. Foreign keys establish relationships at the database level. In order to take advantage of those relationships at the model level, we need to add our `has_many` and `belongs_to` methods to our models.

module-2/backend/week_one.md
```diff
@@ -5,30 +5,123 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+ station = Station.new(name: "Santa Barbara 1")
+ station.persisted?
+      => false
+ station.save
+ station.persisted?
+     => true
+     
+  station_2 = Station.create(name: "San Jose")
+  station.persisted?
+     => true
```
s-espinosa: This is an awesome way to illustrate this difference!

PR #176, opened 2017-01-31
Update week_one.md

scmountain:
@s-espinosa @Carmer


module-2/backend/week_one.md
```diff
@@ -5,30 +5,58 @@ Fork this respository. Answer the questions to the best of your ability. Try to
 Note: When you're done, submit a PR. 
 
 1. List the five common HTTP verbs and what the purpose is of each verb.
+get, puts, delete, post, patch
 2. What is Sinatra?
+sinatra is the framework we use to create web apps
 4. What is MVC?
+model view controller
```
s-espinosa: Yup! Recommend you also be familiar with the responsibilities of each of these pieces.

module-2/backend/week_one.md
```diff
@@ -5,30 +5,58 @@ Fork this respository. Answer the questions to the best of your ability. Try to
 Note: When you're done, submit a PR. 
 
 1. List the five common HTTP verbs and what the purpose is of each verb.
+get, puts, delete, post, patch
 2. What is Sinatra?
+sinatra is the framework we use to create web apps
 4. What is MVC?
+model view controller
 5. Why do we follow conventions when creating our actions/path names in our Sinatra routes?
+allows order to be maintained, and read-ability to outside developers
```

module-2/backend/week_one.md
```diff
@@ -5,30 +5,58 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
 5. Why do we follow conventions when creating our actions/path names in our Sinatra routes?
+allows order to be maintained, and read-ability to outside developers
 6. What types of variables are accessible in our view templates without explicitly passing them?
+instance vars
 7. Given the following block of code, how would I pass an instance variable `count` with a value of `1` to my `index.erb` template?
   
   ```ruby
   get '/horses' do
+    @count = 1
+    @name = "mr. ed"
```
s-espinosa: This is another instance variable. In order to use a local, we need to use slightly different syntax.

```ruby
get '/horses' do
  @count = 1
  name = "Mr. Ed"
end
```

This would give us access to the local variable `name` in the view. If we wanted to, we could also skip the assignment and still have access to the local variable `name`:

```ruby
get '/horses' do
  @count = 1
  erb :index, locals: {name: "Mr. Ed"}
end
```

module-2/backend/week_one.md
```diff
@@ -5,30 +5,58 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
     erb :index
   end
   ```
 
 9. What's the purpose of ERB?
+allows us to write html with ruby injects
```

module-2/backend/week_one.md
```diff
@@ -5,30 +5,58 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+test the several phases without risk of destroying your main data
 11. What's responsive design?
+allows sites to be made to respond to a devices screen size
 12. What is CRUD and why is it important?
+create read update delete, it is the common path creating convention for "building" apps
 13. What does HTTP stand for? 
+hyper text transfer protocol
 14. What are the two ways to interpolate Ruby in an ERB view template? What's the difference between these two ways?
+<%= for displaying the result of the code
+<% just runs the code no need for display
```

module-2/backend/week_one.md
```diff
@@ -5,30 +5,58 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
 15. What's an ORM?
+object relationship mapping
```
s-espinosa: Yup! What's that really mean, though? What is the responsibility of an ORM?

module-2/backend/week_one.md
```diff
@@ -5,30 +5,58 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
 16. What's the most commonly used ORM in ruby (Sinatra & Rails)?
+active record
 17. Let's say we have an application with restaurants. There are seven verb + path combinations necessary to provide full CRUD functionality for our restaurant application. List each of the seven combinations, and explain what each is for.
+get '/' gets the root
```
s-espinosa: Yup, but doesn't deal with the restaurant resource.

module-2/backend/week_one.md
```diff
@@ -5,30 +5,58 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+get '/restaurant' gets restaurant page
+post '/restaurant' displays that view erb
+get '/restaurant/:id/edit' gets the edit erb
+put '/restaurant/:id' edits the info and saves it, redirecting you to id page
+get '/restaurant/:id gets the individual page /
+delete '/horses/:id DESTROYS!!!!!!!!
```
s-espinosa: A few notes -

* the base part of these URLs (`/restaurants`) should be plural for all routes.
* looks like you're missing a GET '/restaurants/new', which displays a form to create a new restaurant.
* POST '/restaurants' creates a new restaurant, and isn't really responsible for displaying a view. It will generally redirect to an index or a show page.

module-2/backend/week_one.md
```diff
@@ -5,30 +5,58 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
 18. What's a migration? 
+where you bring the data base infor to your application
 19. When you create a migration, does it automatically modify your database?
+no
```
s-espinosa: Right! Make sure you know what command you need to run to change your database.

module-2/backend/week_one.md
```diff
@@ -5,30 +5,58 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
 20. How does a model relate to a database?
+model allows the db to exist for manipulation
```
s-espinosa: Be careful here. A database can exist without our models, but our models *need* the database, and the relationships in the database to manipulate the data in the database. Our models are classes that correspond to tables in our database. Instances of our models represent specific rows in those tables.

module-2/backend/week_one.md
```diff
@@ -5,30 +5,58 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
 21. What's the difference between agile workflow and waterfall method?
+agile you work with change. waterfall gets less client feed back
 22. What is the difference between `#new` and `#create`?
+new is just an instance of that entry, create adds the new item permamently
```
s-espinosa: Yup! One thing to know is that `create` runs both the `new` and the `save` command. 🎉

PR #178, opened 2017-02-05
Week Two Recap Questions

JonathanSR:
Answers for Week Two Questions.

### DIFF
--- a/module-2/backend/week_two.md
+++ b/module-2/backend/week_two.md
@@ -5,6 +5,8 @@ Fork this respository. Answer the questions to the best of your ability. Try to 
 Note: When you're done, submit a PR. 
 
 1. At a high level, what is ActiveRecord? What does it do/allow you to do?
+  ActiveRecord is a library, the model in the MVC structure, which is resposible for representing business data and logic. It provides an ORM framework that is wrapped around a relational database. It allows us to wrap our data in Ruby objects so that we can manipulate them. 
+
 2. Assume you have the following model:
 
 ```ruby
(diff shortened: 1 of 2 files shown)


module-2/backend/week_two.md
```diff
@@ -13,8 +15,21 @@ end
 ```
 
+I can call find, group, select, where. 
+If the methods aren't defined in the class then I am not sure as how we have access to them.
```
s-espinosa: We have access because those methods are defined in ActiveRecord and we're inheriting from that.

module-2/backend/week_two.md
```diff
@@ -13,8 +15,21 @@ end
...
 
 3. Assume that in your database, a team has the following attributes: "id", "name", owner_id". How would you find the name of a team with an id of 4? Assuming your class only included the code from question 2, how could you find the owner of the same team?
+ ```ruby
+  team4 = Team.find(4)
+  team4.name
+end
```
s-espinosa: Don't need an end here.

module-2/backend/week_two.md
```diff
@@ -13,8 +15,21 @@ end
...
+```
+
+```ruby
+  team4 = Team.find(4)
+  team4.owner_id
+  end
```
s-espinosa: Same as above.

Also, this finds the owner id, but not the owner itself. In order to find the owner, we would need to do something like the following:

```ruby
owner_4_id = team4.owner_id
Owner.find(owner_4_id)
```

module-2/backend/week_two.md
```diff
@@ -26,26 +41,61 @@ end
 
 Now how would you find the owner of the team with an id of 4?
 
+```ruby
+  team4 = Team.find(4)
+  team4.owner
```

module-2/backend/week_two.md
```diff
@@ -26,26 +41,61 @@ end
...
+
 2. Name your three favorite ActiveRecord rake tasks and describe what they do.
+rake db:migrate:reset - This runes drop, create and migrate, therfore dropping your database, creating it and migrating the migrations.
+rake db:test:prepare - This setups your test environment.
+rake db:seed - This seeds your database.
+
 4. What can you expect from a group as you begin working together? As you continue working together?
+
 5. What two columns does `t.timestamps null: false` create in our database?
+This creates two time columns, one is a created at and the other is updated at, the updated column changes with every update to that specific row/object.
```

module-2/backend/week_two.md
```diff
@@ -26,26 +41,61 @@ end
...
 7. What case does JSON (and JavaScript) use for multi-word variables?
 8. What case does Ruby use for multi-word variables?
 9. In a database that's holding schools and teachers, what will be the relationship between schools and teachers?
+Schools and teachers is a one to many relationship. A school can have many teachers and a teacher can have one school.
+
 10. In the same database, what will you need to do to create this relationship (draw a schema diagram)?
+The two tables, schools and teachers, would be joined by the primary key of the school that would become the foreign key in the teachers table.
+
 11. Give an example of when you might want to store information besides ids on a join table.
+When attributes are created or become byproducts of joining two tables together.
```

PR #180, opened 2017-02-05
week2 cfu charlotte

CjMoore:


module-2/backend/week_two.md
```diff
@@ -25,30 +32,70 @@ end
 ```
 
-
+<p>
+Answered above. Another way would be to set the team with an id of 4 to a variable and call owner on that, 'team = Team.find(4)' 'team.owner'
```

Above I think we were assuming that there were no changes to the model, which would require some less than ideal code like the following:

```ruby
owner = Owner.find(owner_4_id)
```

module-2/backend/week_two.md
```diff
@@ -25,30 +32,70 @@ end
...
+</p>
 9. Describe the relationship between a foreign key on one table and a primary key on another table.
+<p>
+Using the Team and Owner examples above, the foreign key on the Team table has a value equal to the primary key on the Owner table that relates to the Owner of that particular team. More specifically, lets say we have a team who's name is something like "Wizards" and they're owned by a guy named "Bob". If the instance of "Bob" the owner has a primary key of 1, the instance of "Wizards" on the Team table will have an attribute of owner_id who's value is 1. For the sake of clarity, lets say the relationship between Teams and Owners is such that a team only has one owner but an owner can have many teams. If "Bob" also owns a team named "Dragons", then the instance of "Dragons" on the team table will also have an owner_id attribute who's value is equal to 1.
```

PR #183, opened 2017-02-06
@cmacaulay1220's week 2 understanding check responses. 

cmacaulay:
👍 ⚡️

### DIFF
--- a/module-2/backend/week_two.md
+++ b/module-2/backend/week_two.md
@@ -5,6 +5,7 @@ Fork this respository. Answer the questions to the best of your ability. Try to 
 Note: When you're done, submit a PR. 
 
 1. At a high level, what is ActiveRecord? What does it do/allow you to do?
+ActiveRecord is a library of methods that allows developers to use ruby to interact with the database. This is possible because the ActiveRecord methods are writing the sequel for us. 
 2. Assume you have the following model:
 
 ```ruby
@@ -13,8 +14,21 @@ end
 ```
 
 What are some methods you can call on `Team`? If these methods aren't defined in the class, how do you have access to them?
-
+maximum
+average
+group
+find_or_create_by
+where
+order
+uniq
+pluck
+joins
+=> through Activerecord. 
 3. Assume that in your database, a team has the following attributes: "id", "name", owner_id". How would you find the name of a team with an id of 4? Assuming your class only included the code from question 2, how could you find the owner of the same team?
+team = Team.where(id: 4)
+team.name 
+=> "Team Name"
+team.owner_id.pluck(:name)
 
 4. Assume that you added a line to your `Team` class as follows:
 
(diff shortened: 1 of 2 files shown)


module-2/backend/week_two.md
```diff
@@ -13,8 +14,21 @@ end
...
+order
+uniq
+pluck
+joins
+=> through Activerecord. 
 3. Assume that in your database, a team has the following attributes: "id", "name", owner_id". How would you find the name of a team with an id of 4? Assuming your class only included the code from question 2, how could you find the owner of the same team?
+team = Team.where(id: 4)
+team.name 
+=> "Team Name"
+team.owner_id.pluck(:name)
```
s-espinosa: Can't use pluck here! `team.owner_id` is going to return a number (the id of the owner), and we can only call `pluck` on ActiveRecord objects. Instead we need to first find the owner itself.

```ruby
Owner.find(owner_4_id)
```

module-2/backend/week_two.md
```diff
@@ -23,22 +37,35 @@ class Team << ActiveRecord::Base
   belongs_to :owner
 end
 ```
-
+team.owner.name
```
s-espinosa: 👍 This would find the owner's name. We're technically just looking for the owner object, but it looks like you've got the right idea.

module-2/backend/week_two.md
```diff
@@ -23,22 +37,35 @@ class Team << ActiveRecord::Base
...
+Call on the singular of that column and have access to its attributes. 
 7. In a database that's holding students and teachers, what will be the relationship between students and teachers? Draw the schema diagram.
+
+Posting load code in a comment.
 8. Define foreign key, primary key, and schema.
+Schema - This holds the basic structure of your data-base. Migrations determine the makeup of the schema. Where tables live. 
+Primary Key - Unique identifier that every record is created with. 
+Foreign key - Link to another table through the unique identifier of a specific record in the second table. 
 9. Describe the relationship between a foreign key on one table and a primary key on another table.
+They are the same?
```
s-espinosa: Yup! I might add something to indicate that they define the relationship between the two tables.

module-2/backend/week_two.md
```diff
@@ -23,22 +37,35 @@ class Team << ActiveRecord::Base
...
+
+Posting load code in a comment.
 8. Define foreign key, primary key, and schema.
+Schema - This holds the basic structure of your data-base. Migrations determine the makeup of the schema. Where tables live. 
+Primary Key - Unique identifier that every record is created with. 
+Foreign key - Link to another table through the unique identifier of a specific record in the second table. 
 9. Describe the relationship between a foreign key on one table and a primary key on another table.
+They are the same? 
 10. What are the parts of an HTTP response?
+Client --> Request --> Routes --> Controller --> Model --> Database
```
s-espinosa: This is how that HTTP response is created, but the *parts* of the HTTP response are the headers and body (the status code is the first line of the headers). Still, nice to practice laying out the MVC pattern.

PR #187, opened 2017-02-06
Answers for week 2

edilenedacruz:
Only the basic questions were answered.


module-2/backend/week_two.md
```diff
@@ -13,8 +14,14 @@ end
 ```
 
+  We can call on Team any ActiveRecord method even tough they are not describe.
```
s-espinosa: Yes! But what are some of those methods?

module-2/backend/week_two.md
```diff
@@ -13,8 +14,14 @@ end
...
 
 3. Assume that in your database, a team has the following attributes: "id", "name", owner_id". How would you find the name of a team with an id of 4? Assuming your class only included the code from question 2, how could you find the owner of the same team?
+  
+  Team.find(4).name
+  
+  Team.find(4).owner_id
+  Owner.find(ownder_id)
```
s-espinosa: Doesn't look like we've defined `owner_id` as a local variable anywhere in this code. You'd need to save that to a local to be able to use it here in this last line. For example:

```ruby
Team.find(4).name
Owner.find(owner_4_id)
```

module-2/backend/week_two.md
```diff
@@ -27,12 +34,24 @@ end
 Now how would you find the owner of the team with an id of 4?
```
s-espinosa: Recommend giving this a try! Let us know if you'd like to discuss.

module-2/backend/week_two.md
```diff
@@ -27,12 +34,24 @@ end
...
 8. Define foreign key, primary key, and schema.
+  foreign key = a foreign key references a row at another table, it will have that tables name followed by 'id'
+  primary key = it's a unique identifier for a table's row/object
+  schema = it's the blueprint of the tables
+  
 9. Describe the relationship between a foreign key on one table and a primary key on another table.
+  The foreign key corresponds to the primary key from another table. 
+  
 10. What are the parts of an HTTP response?
+  Start line, header and body.
```
s-espinosa: Think that first part maybe you were thinking of status? An HTTP response has headers and a body and the first header line is a status code (e.g. 404 - not found, 200 - ok, 503 - server error, etc).

PR #212, opened 2017-03-07
Update week_one.md

AliSchlereth:
Add review content as well as self assessment to the first weekly check in for M3BEE 
What do you think @s-espinosa ?

### DIFF
--- a/module-3/backend/week_one.md
+++ b/module-3/backend/week_one.md
@@ -4,8 +4,23 @@ Fork this respository. Answer the questions to the best of your ability. Try to 
 
 Note: When you're done, submit a PR. 
 
-1. What is `json` and why is it important?
+1. What is `json`, what does it stand for, and why is it important?
 2. What's the difference between `joins` and `includes` in ActiveRecord?
 3. What's an API?
 4. How do we test an internal API (in general)?
-5. What are two different ways to customize your `json`?
+5. What are two different ways to customize your `json`?  
+
+#### Review  
+6. What is an ORM, what does it stand for, and why is it helpful?  
+7. How do you create a record in the following table in SQL? In Active Record?   
+   (Users table with columns first_name, last_name, email, and age)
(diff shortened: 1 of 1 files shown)


module-3/backend/week_one.md
```diff
@@ -4,8 +4,23 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
 3. What's an API?
 4. How do we test an internal API (in general)?
-5. What are two different ways to customize your `json`?
+5. What are two different ways to customize your `json`?  
+
+#### Review  
+6. What is an ORM, what does it stand for, and why is it helpful?  
+7. How do you create a record in the following table in SQL? In Active Record?   
+   (Users table with columns first_name, last_name, email, and age)
+8. Given an array numbers = [1,2,3,4,5], find the sum of the doubles of all the numbers.
```
s-espinosa: Maybe add something to explicitly ask them to do this in Ruby?

module-3/backend/week_one.md
```diff
@@ -4,8 +4,23 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
-5. What are two different ways to customize your `json`?
+5. What are two different ways to customize your `json`?  
+
+#### Review  
+6. What is an ORM, what does it stand for, and why is it helpful?  
+7. How do you create a record in the following table in SQL? In Active Record?   
+   (Users table with columns first_name, last_name, email, and age)
+8. Given an array numbers = [1,2,3,4,5], find the sum of the doubles of all the numbers.  
+
+#### Self Assessment
```
s-espinosa: Interesting! I think I like this, but let's discuss next week. Think it would be great if we could catch up on Wednesday and re-calibrate based on where we think things stand at that point.

module-3/backend/week_one.md
AliSchlereth: Yes, good point!

module-3/backend/week_one.md
AliSchlereth: Sounds like a solid plan to me.

PR #213, opened 2017-03-17
add BE-Mod 3 week 1 responses for casey macaulay

cmacaulay:


module-3/backend/week_one.md
```diff
@@ -5,7 +5,22 @@ Fork this respository. Answer the questions to the best of your ability. Try to
 Note: When you're done, submit a PR. 
 
 1. What is `json` and why is it important?
+```
+JSON is what APIs are generally displayed as (or XML). It takes the information from a database and outputs it so that it can be consumed by a front end or someone else who wants the data. The controller is where json is rendered. 
+```
 2. What's the difference between `joins` and `includes` in ActiveRecord?
+```
```
s-espinosa: Includes works very much like joins, but also pre-loads some of the data on the other table to prevent multiple queries where possible.

cmacaulay: @s-espinosa I will look into it! Do you happen to know of any resources that have a good visual representation of these methods? I think that would be most helpful for how I learn. If I happen to find any I'll pass them along to you!

PR #216, opened 2017-03-17
Complete CFU for week 1 mod 3

cews7:


module-3/backend/week_one.md
```diff
@@ -5,22 +5,30 @@ Fork this respository. Answer the questions to the best of your ability. Try to
 Note: When you're done, submit a PR. 
 
 1. What is `json`, what does it stand for, and why is it important?
+*Javascript object rotation. It is a way of taking data from one app and serving it to another. We could also use XML to do this, but that would be painful.*
```
s-espinosa: I can't speak to how painful it would be to serve XML, but think it's important to be flexible. We taught JSON, many of the APIs you consume will use JSON, but I don't think it's unreasonable to imagine a case where someone would ask you to implement XML for some reason or another.

module-3/backend/week_one.md
```diff
@@ -5,22 +5,30 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
 2. What's the difference between `joins` and `includes` in ActiveRecord?
+*joins is used to join two tables together based on a single attribute.*
```
s-espinosa: `.includes` can do this as well, but will also preload related data to avoid additional queries.

PR #217, opened 2017-03-19
add Jonathan Kidd responses to CFU mod 2, week 1

jk1dd:
Added answers for most questions.

Number 6 didn't know at all, and sure there are some other gaps throughout.


bethsebian:
Looks great, @jk1dd!

For `What types of variables are accessible in our view templates without explicitly passing them?`, just looking for `instance variables`.

For `17. Seven CRUD verb + path combinations`, the one you're missing is `/restaurants/#id` which will let you view a specific restaurant.

PR #225, opened 2017-03-20
Natalie's completed Update week_one.md

NatBee:
Completed week 1 review.


bethsebian:
Looks great @NatBee!

For `7`, looking for something like this.

```
get '/horses' 
  @count = 1
  erb :index, :locals => {name: "Mr. Ed"}
end

For `11. What's responsive design?` looking for: design that can adjust based on the browser size and other variables.

For `12. What is CRUD and why is it important?`, looking for: CRUD covers all the ways we will need to interact with a database: creating, updating, deleting and viewing.

PR #229, opened 2017-03-20
Update week_one.md

josh-works:


bethsebian:
Looks good @josh-works!

For `1. List the five common HTTP verbs and what the purpose is of each verb`, `PUT` and `POST` both add to the database, but `POST` is designed to add a new record to the database, while `PUT` updates an existing record.

For `6. What types of variables are accessible in our view templates without explicitly passing them?`, looking for "Instance Variables."

For `8. In the same code block, how would I pass a local variable `name` with a value of `Mr. Ed` to the view?` looking for (note: this illustrates two different ways to prepare data for the view):

get ‘/horses/ do 
	@count = 1
	erb :index, locals => {:name = “Mr. Ed”}
end

For `10. Why do I need a development AND test database?`: Having a separate testing database  gives you a controlled environment in which you know exactly what data you have at any given time, and thus can successfully test around.

For `17. The seven CRUD verb + path combinations`:

- get '/restaurants' => show all of the restaurants (list) - YOU GOT THIS ONE 👍 
- get '/restaurants/:id'  => show a specific restaurant - YOU GOT THIS ONE 👍 
- get '/restaurants/new' => forum to add a new restaurant - ADDED
- POST '/restaurants' => forum to add a new restaurant - ADDED
- get '/restaurants/:id/edit' => show the page to edit a specific restaurant - YOU GOT THIS ONE 👍
- put '/restaurants/:id' => the above page would post to the DB to update the restuarant info - YOU GOT THIS ONE 👍
- delete '/restaurants:id' => delete the restaurant - YOU GOT THIS ONE 👍
- get '/' => show the application root - NOT A THING
-

PR #230, opened 2017-03-20
Erin's update/review - mod2 week one learnings

somedayrainbows:
my answers inline


module-2/backend/week_one.md
```diff
@@ -1,34 +1,64 @@
...
   get '/horses' do
+    @count = 1
     erb :index
   end
   ```
-
+   ```ruby
+  get '/horses' do
+     @name = "Mr. Ed"
```
somedayrainbows: dangit! i misread this initially! to pass a local variable, you'd do something like:

get '/horses' do
  name = "Mr. Ed"
end

bethsebian:
Looks great @somedayrainbows!

For `19. When you create a migration, does it automatically modify your database?`, just looking for: "You also need to run a migration for it to take effect"

PR #232, opened 2017-03-20
Joseph Glass homework answers

glassjoseph:


bethsebian:
This all looks great, @glassjoseph! Especially impressive because you're a week ahead!

PR #233, opened 2017-03-20
Sam's diagnostic

samlandfried:
I did this two times =(


bethsebian:
Looks good, @samlandfried !

For `2. What is Sinatra?`, worth noting that it's a server with access to a database.

For `5. Why do we follow conventions when creating our actions/path names in our Sinatra routes?`, worth noting that it also helps actual humans figure out what's going on. 👍

PR #235, opened 2017-03-20
Bschwartz week 1 check for understanding

bschwartz10:


bethsebian:
Looks good @bschwartz10!

PR #238, opened 2017-03-20
Week 1 check

wewert:


bethsebian:
Hey @wewert! Great start here! I think you'd get some benefit from reviewing these questions again and doing some research to look up the answers. Focus specifically on questions 6-9 and 15-20 -- I think you'll get a lot of value out of that

PR #244, opened 2017-03-26
Colleen week 2 checks for understanding

caward12:


bethsebian:
Nice work @caward12! 👍

PR #251, opened 2017-03-27
Update week_two.md

markyv18:


bethsebian:
Pretty solid @markyv18!

On 3, looking for the actual object, thus `Owner.find(4)`
On 4, looking for `Owner.find(4).owner`

markyv18:
got it, thanks :)

PR #257, opened 2017-03-27
Lauren Oliveri - Mod 2 Checks for Understanding

lao9:
For both week one and two since I was behind there!


bethsebian:
Really strong work here, @lao9!

PR #258, opened 2017-03-27
Jack week_two.md

jwpincus:


bethsebian:
Looks good @jwpincus!

For 8, also looking for: path, status code, verb, header(s).

PR #263, opened 2017-03-27
Update week_two.md

stovermc:


bethsebian:
Looks good @stovermc!

For 2, also: delete.

For 8, think also: path, status code, verb, headers

PR #264, opened 2017-03-28
add wk two diagnostic

mariastenquist:
I'm still struggling with the AR relationships and using AR methods correctly. Also getting bogged down with which elements are plural, singular, and which order (e.g. do I call the table that something belongs_to first, and then has_many?). Also, with the bash commands for rails--the syntax is pretty specific (e.g. spacing) and still am trying to wrap my head around order & syntax.


Carmer:
Thanks @mariastenquist . We have some AR practice setup this week. I think these are good questions overall. At some time this week lets see if we can talk about this.

PR #266, opened 2017-03-31
Jack P Week Three

jwpincus:
This exercise was pretty straight forward. I did use google to get syntax right, but otherwise knew the concepts.


Carmer:
good work @jwpincus

PR #267, opened 2017-04-01
Joseph Glass - Week 3 Recap

glassjoseph:
I feel strong on most of these concepts, though some of the syntax details aren't yet in long-term memory.


Carmer:
Good work @glassjoseph . I think the long term memory will come with a little more time.

PR #268, opened 2017-04-02
Colleen week three checks for understanding

caward12:


Carmer:
👍 Good work @caward12

PR #271, opened 2017-04-03
Lauren Oliveri Week 3 CFU

lao9:
Sorry for the delay! Here's my Week 3 Check for Understanding.


Carmer:
👍 Good answers. Good work being thorough.

PR #279, opened 2017-05-07
Edilene

edilenedacruz:
This completes the CFU week one.


module-4/backend/week_one.md
```diff
@@ -1,21 +1,63 @@
...
 4. What is the syntax for invoking a function?
+
+  ```javascript
+  function();
+  ```
+
 5. What's the difference between `==` and `===` in JavaScript?
-6. What's the difference between asynchronous and synchronous JavaScript? 
+
+  `==` is used for checking equality and `===` is checking for a true/false result
```
case-eee: type conversion is done with `==` and with `===`, there is not type conversion, it's just checking for equality. 

`'7' == 7` will return true
`'7' === 7` will return false

module-4/backend/week_one.md
```diff
@@ -1,21 +1,63 @@
...
+
+6. What's the difference between asynchronous and synchronous JavaScript?
+
+  In asynchronous JS the code is executed without an specific order and while long, slow code is being executed, a smaller and faster one will execute a the same time; in synchronous JS, the code is executed in order and the following command/code will only be executed after the prior code is finished.
+
 7. What's a callback function and what are some reasons when we use/need callback functions?
+
+  A callback function is a function that is passed to another function as a parameter. The callback function can be executed inside that other function.
```
case-eee: 👍

module-4/backend/week_one.md
```diff
@@ -1,21 +1,63 @@
...
+
 8. What's the biggest difference between a promise and a callback?
+
```
case-eee: i suggest doing some research with this! it's important to know both of these and what they can do for you. comparing/contrasting may not be as important

edilenedacruz: Ooops. I forgot to go back to it. :sad_panda:

PR #313, opened 2017-05-15
M4: Week two cfu

Ryanspink1:


module-4/backend/week_two.md
```diff
@@ -0,0 +1,56 @@
...
+
+- server.js acts like routes, sends to controller --> model --> database --> model --> controller
+
+6. How do you execute raw SQL in node?
+
+using knex: knex.raw(SQL)
+
+7. What is CORS?
+
+-CORS is cross origin resource sharing. It restricts access to resources.
```
case-eee: 👍

module-4/backend/week_two.md
```diff
@@ -0,0 +1,56 @@
...
+
+- functions replaced by hash rockets
+
+4. What's the deal with semi-colons in JavaScript?
+
+- not necessary but helpful for reading. Apparently people are quite opinionated on this.
+
+5. How are you using the MVC design pattern in your Quantified Self project?
+
+- server.js acts like routes, sends to controller --> model --> database --> model --> controller
```
case-eee: 👍

module-4/backend/week_two.md
```diff
@@ -0,0 +1,56 @@
+## Week Two - Module 4 Recap
+
+
+1. What is Webpack and why is it useful?
+
+- It's essentially a bundler for JS. Allows you to pull many different resources together under one, or multiple codebases.
+
+2. When do you want to use event delegation?
```
case-eee: this is why sometimes we need to add an event listener to a parent rather directly on the child - i think this may have been why you had trouble with the delete events!

PR #387, opened 2017-08-21
squee week one answers

squeemishly:


module-4/backend/week_one.md
```diff
@@ -5,18 +5,79 @@ Fork this respository. Answer the questions to the best of your ability. Try to
 Note: When you're done, submit a PR. 
 
 1. What's the most useful thing you learned from completing the intermission week work?
+```
+That I actually really enjoy JS. Mostly, the intermission work was review of stuff we'd done in Mod 3.
```
case-eee: wooooo enjoying JS 👍

module-4/backend/week_one.md
```diff
@@ -5,18 +5,79 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+```
+
 4. What is the syntax for invoking a function?
+```
+functionName();
+```
+
 5. What is CORS?
+```
+CORS is a method to share resources between domains, e.g. a call to an API in one domain to get information for another domain.
```
case-eee: absolutely

module-4/backend/week_one.md
```diff
@@ -5,18 +5,79 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+```
+
 6. How do we enable CORS?
+```
+... Mostly, I just added the methods I wanted access to in config/application.rb... Umm... Is there a gem? Do you just ask the internet and blindly follow their rules? Surly there's docs somewhere... ;)
```
case-eee: yup, we're using a gem in our Rails backend. we'll need to handle it in a different way when we create our node backend 👍

module-4/backend/week_one.md
```diff
@@ -5,18 +5,79 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+```
+
 7. What's `this` in JavaScript?
+```
+huge.
+Also, it's a much bigger version of 'self' in Ruby. Depending on where you call 'this', it could be any of a myriad of different content. 'this' could be the window or it could be a particular line you are clicking on. It's scope changes depending on where you are in your code.
```
case-eee: great overview

module-4/backend/week_one.md
```diff
@@ -5,18 +5,79 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+```
+
 8. What is Webpack and why is it useful?
+```
+Webpack is a way of organizing and utilizing your resources (e.g. all your .scss and .js files) in your app. Instead of individually including each file on each page that requires it, webpack will condense them all down and do the including for you. Nifty.
+```
+
 9. When/why do you want to use event delegation?
+```
+When you have resources that are being dynamically loaded on the page. For example, if you have content that loads after it makes an API call, that content will not be available on $(document).ready. If you need to access that content after the document loads, you need to bind it to another element that is found on $(document).ready and call it from there.
```
case-eee: absolutely 🎉

module-4/backend/week_one.md
```diff
@@ -5,18 +5,79 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+```
+
 10. What is a callback function?
+```
+A callback function is a function that is passed to another function as a parameter and then is executed inside the original function. Function. Function. Function...
+```
+
 11. What's `npm` and what do we use it for?
+```
+npm = Node Package Manager. It's basically the gem of JS. It allows developers to share packets of code to simplify their coding in the future.
```
case-eee: a package is similar to a gem and npm is similar to bundler 👍

module-4/backend/week_one.md
```diff
@@ -5,18 +5,79 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+
 11. What's `npm` and what do we use it for?
+```
+npm = Node Package Manager. It's basically the gem of JS. It allows developers to share packets of code to simplify their coding in the future. 
+```
 
 #### Review  
 12. What's the MVC design pattern? Describe each part of MVC.
+```
+Models - access the database, find specific information, do most of the logicing, all your fancy methods are belong to us
```
case-eee: great overview

module-4/backend/week_one.md
```diff
@@ -5,18 +5,79 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+Views - render the content they're told to by controllers, do almost no logicing, none of your fancy methods belong in here
+Controllers - do minimal logicing, act as the go between for views and models, your fancy methods might be called in here
+```
+
 13. What are a few ways to optimize a Rails application?
+```
+Keep all logic in the Model
+Use ActiveRecord as often as possible to manage and access information in your database
```
case-eee: use the query methods that are most effective; possibly cache content; avoid n+1 ActiveRecord queries

module-4/backend/week_one.md
squeemishly: Hey @case-eee, I don't know that we ever explicitly covered this... In fact, I was thinking about this over the weekend... I understand the principle of using tokens to limit access to certain requests (e.g. delete table users), but I don't understand the practical side of this. I don't see a lesson on this coming up. Will our class be able to look at this before graduation? If not, would you be willing to help me figure it out one on one?

case-eee: @squeemishly we can definitely talk about it! we don't have a class on it specifically. we briefly touch on it when you created your node backend and handled requests coming from your separate front end application. let me know when you want to chat more about it!

PR #391, opened 2017-08-21
Week One CFU - 1706 BEE - Aaron Hursh

amhursh:
@s-espinosa @icorson3

### DIFF
--- a/module-2/backend/week_one.md
+++ b/module-2/backend/week_one.md
@@ -1,34 +1,111 @@
 ## Week One - Module 2 Recap
 
-Fork this respository. Answer the questions to the best of your ability. Try to answer them with limited amount of external research. These questions cover the majority of what we've learned this week (which is a TON!). 
+Fork this respository. Answer the questions to the best of your ability. Try to answer them with limited amount of external research. These questions cover the majority of what we've learned this week (which is a TON!).
 
-Note: When you're done, submit a PR. 
+Note: When you're done, submit a PR.
 
 1. List the five common HTTP verbs and what the purpose is of each verb.
+
+  get - This is used to read a record
+  post - Used to create a record
+  put - Used to update a record
+  patch - Also used to update a record??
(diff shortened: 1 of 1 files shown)


module-2/backend/week_one.md
```diff
@@ -1,34 +1,111 @@
...
 
-Note: When you're done, submit a PR. 
+Note: When you're done, submit a PR.
 
 1. List the five common HTTP verbs and what the purpose is of each verb.
+
+  get - This is used to read a record
+  post - Used to create a record
+  put - Used to update a record
+  patch - Also used to update a record??
```
s-espinosa: Yup! PUT/PATCH are generally used interchangeably, though PUT is technically to fully replace a resource while PATCH is to update some attributes.

module-2/backend/week_one.md
```diff
@@ -1,34 +1,111 @@
...
+
+  This is the model-view-controller structure that we use in Sinatra (and maybe Rails?) that is a sort of a way to maintain the spirit of single responsibility in our apps. We break things into a controller, model(s), and views.
+
 5. Why do we follow conventions when creating our actions/path names in our Sinatra routes?
+
+  I would imagine it is to make our lives (as programmers) easier and make it less likely for bugs and other errors to cripple apps.
+
 6. What types of variables are accessible in our view templates without explicitly passing them?
+
+  anything in the params.
```
s-espinosa: Those are accessible in our controller, but *instance* variables are available in our views.

module-2/backend/week_one.md
```diff
@@ -1,34 +1,111 @@
...
     erb :index
   end
   ```
 
+
+  ```ruby
+  get '/horses' do
+    @name = 'Mr. Ed'
+    erb :index
```
s-espinosa: Using a local variable we would change these two lines to:

```ruby
name = "Mr. Ed"
```

PR #399, opened 2017-08-21
Randy Springer 1706 BE CFU Week 1

randy-darsh:


module-2/backend/week_one.md
```diff
@@ -4,31 +4,90 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+  get '/horses' do
+  :locals {:mr.ed => mr.ed}
+  end
+  ```
+ 
 9. What's the purpose of ERB?
+So you can render ruby in HTML
+
 10. Why do I need a development AND test database?
+Development you can change as you go. Test would be like a fixture that you can test for specific things.
```
s-espinosa: I wouldn't consider the test database a fixture exactly. Fixtures create data that we can use in our tests. Our test database is a database that exists separate from production/dev that allows us to create/destroy data that we can test without impacting those databases.

module-2/backend/week_one.md
```diff
@@ -4,31 +4,90 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+
 17. Let's say we have an application with restaurants. There are seven verb + path combinations necessary to provide full CRUD functionality for our restaurant application. List each of the seven combinations, and explain what each is for.
-18. What's a migration? 
+READ - GET - view all<br>
+READ - GET - view one<br>
+CREATE - GET - form for new thing<br>
+CREATE - POST - creating the new thing<br>
+UPDATE - GET - form for editing the thing<br>
+UPDATE - PUT/PATCH - updating the thing<br>
+DELETE - DESTROY - deleting the thing<br>
```
s-espinosa: Need the paths (e.g. `/restaurants`) as well!

module-2/backend/week_one.md
```diff
@@ -4,31 +4,90 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+
+
+18. What's a migration?
+Where your table is created and attributes are assigned.
+
 19. When you create a migration, does it automatically modify your database?
+No. You create the migration. Then inside you put your validations and relationships then you use rake db:migrate to update your database schema.
```
s-espinosa: Validations/relationships will go in your model. The column/table names go into your migration.

PR #406, opened 2017-08-21
Aaron Week 1 Checks for Understanding

MrAaronOlsen:


module-4/backend/week_one.md
```diff
@@ -5,18 +5,43 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
 3. What are some tools you need in order to unit test your JavaScript?
+Mocha at the least, and Chai for better asserts.
 4. What is the syntax for invoking a function?
+```javascript
+function() <- needs the parantheses, function just returns the actual function, which is useful for sending callbacks etc...
+```
 5. What is CORS?
+Cross-Origin Resource Sharing. My understanding is that it allows one web origin to access data from another web origin.
 6. How do we enable CORS?
+I have no idea. Yet.
```
laurenfazah: Server-side config will set this up to restrict requests to API. It's enabled automatically client-side by Chrome

module-4/backend/week_one.md
```diff
@@ -5,18 +5,43 @@ Fork this respository. Answer the questions to the best of your ability. Try to
 Note: When you're done, submit a PR. 
 
 1. What's the most useful thing you learned from completing the intermission week work?
+
```
laurenfazah: :(

module-4/backend/week_one.md
```diff
@@ -5,18 +5,43 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
 4. What is the syntax for invoking a function?
+```javascript
+function() <- needs the parantheses, function just returns the actual function, which is useful for sending callbacks etc...
+```
 5. What is CORS?
+Cross-Origin Resource Sharing. My understanding is that it allows one web origin to access data from another web origin.
 6. How do we enable CORS?
+I have no idea. Yet.
 7. What's `this` in JavaScript?
+This is similar to 'self' in Ruby, but with some intersting scoping differences. In a nutshell, it is a reference to the object 'this' is wrapped in. Being inside a for loop or forEach method, for example, 'this' will reference the loop. This starts to get weird inside of a class. If you're in declared method in the class this references the entire class. As soon as you enter another function nested inside of that fuction 'this' will now reference that function. You can bind 'this' to a function by attatching {}.bind(this) to the end of it. This will scope 'this' to the object directly outside of the function.
```
laurenfazah: Great explanation

module-4/backend/week_one.md
```diff
@@ -5,18 +5,43 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
 
 #### Review  
 12. What's the MVC design pattern? Describe each part of MVC.
+Model, View, Controller.
+The model accesses your database or acts as a non-database object or PORO (plain ol' ruby object).
+The view renders your html. It will need access to objects to render dynamic pages, but 99% of logic should be kept out.
+The controller actually does a few things. It acts as a crossroads for your RESTful routes, but then also collects data for access by the view. The access data part _should_ be minimal. If a lot of data needs to be collected that responsibility should be moves to a presenter.
+
 13. What are a few ways to optimize a Rails application?
+- Don't write it in Rails.
```
laurenfazah: I'd extrapolate your suggestion to not use Rails (it can be a serious suggestion, just back it up)

module-4/backend/week_one.md
```diff
@@ -5,18 +5,43 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+- Reduce or optimize DB calls.
+- Move DB calls to AJAX. This will render pages faster but will also give you more flexibility with how the page is rendered.
+- Partials can be handy to prevent non-changing views from constantly being re-rendered.
+- Cache data. Large database calls, images and media, large js files, etc.
+
 14. What's a background worker? When would we want to use a background worker?
+A background worker is a function or application that runs in the... _Background!_ A background worker is an asynchronous function that can do things while the main app continues to function for the user. A worker will generally occupy a processor on the server side. It could be a scheduled worker that fires up every now and then. It could be triggered from the application, such as sending E-mail. It could also be setup as a watcher waiting for something to happen, such as messages.
```
laurenfazah: 👍

PR #408, opened 2017-08-22
Update week_one.md

limsammy:


module-2/backend/week_one.md
```diff
@@ -5,30 +5,102 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+
+MVC stands for Model, View, Controller, a systematic approach to object oriented web design
+
 5. Why do we follow conventions when creating our actions/path names in our Sinatra routes?
+
+In order to prepare for all the things Rails offers us, and that web conventions are normally the same across all frameworks
+
 6. What types of variables are accessible in our view templates without explicitly passing them?
+
+Params
```
s-espinosa: We were looking for instance variable (consistent with your answer to 7). I would aim to stay away from accessing params directly in the view.

module-2/backend/week_one.md
```diff
@@ -5,30 +5,102 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
   ```ruby
   get '/horses' do
+    @count = 1
     erb :index
   end
   ```
 
+
+@name = "Mr. Ed"
```
s-espinosa: The `@` makes this an instance variable. To access a local variable you need to pass `locals` as part of the options hash.

```ruby
get '/horses' do
  name = "Mr. Ed"
end
```

module-2/backend/week_one.md
```diff
@@ -5,30 +5,102 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+
 9. What's the purpose of ERB?
+
+It's essentially html markdown in which ruby variables can be passed
```
s-espinosa: And ruby statements evaluated.

module-2/backend/week_one.md
```diff
@@ -5,30 +5,102 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+
 10. Why do I need a development AND test database?
+
+The test database is for spec tests, it cleans itself
```
s-espinosa: It will clean itself if we implement a teardown method (which we're doing by using database cleaner).

module-2/backend/week_one.md
```diff
@@ -5,30 +5,102 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+
+Create, Read, Update, Delete, it is the foundation for most database based applications.
+
 13. What does HTTP stand for? 
+
+HyperText Transfer Protocol
+
 14. What are the two ways to interpolate Ruby in an ERB view template? What's the difference between these two ways?
+
+<%= and <%. The former is for a single line, the latter a block.
```
s-espinosa: The former will be rendered in the HTML when evaluated; the latter will not.

module-2/backend/week_one.md
```diff
@@ -5,30 +5,102 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+ActiveRecord
+
 17. Let's say we have an application with restaurants. There are seven verb + path combinations necessary to provide full CRUD functionality for our restaurant application. List each of the seven combinations, and explain what each is for.
-18. What's a migration? 
+
+* /restaraunts - GET
+* /restaraunt/:id - GET
+* /restaraunt/:id/edit - PATCH
+* /restaraunts/new - PUTS & POST
+* /restaraunts/:id/delete - DELETE
```
s-espinosa: Missing 2. Also, explain what each is used for.

module-2/backend/week_one.md
```diff
@@ -5,30 +5,102 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+
+A modification to the database
+
 19. When you create a migration, does it automatically modify your database?
+
+No, one must run `rake db:migrate`
+
 20. How does a model relate to a database?
+
+A model is the singular noun of a table
```
s-espinosa: Model isn't just the singular noun, it's the class that we use to access the information in our database.

PR #417, opened 2017-08-24
Sarah Kirk CFU week one

case-eee:


module-4/backend/week_one.md
```diff
@@ -5,18 +5,69 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+Right click the site you want to enable CORS for and go to Properties.
+Change to the HTTP Headers tab.
+In the Custom HTTP headers section, click Add.
+Enter Access-Control-Allow-Origin as the header name.
+Enter * as the header value.
+Click Ok twice.
+
 7. What's `this` in JavaScript?
+
+'this'is the thing that owns the javascript code. This is usually the object when used in the object itself.
```
case-eee: 👍

module-4/backend/week_one.md
```diff
@@ -5,18 +5,69 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+Enter * as the header value.
+Click Ok twice.
+
 7. What's `this` in JavaScript?
+
+'this'is the thing that owns the javascript code. This is usually the object when used in the object itself. 
+
 8. What is Webpack and why is it useful?
+
+webpack is a module bundler. It is used in conjunction with selenium to make sure our tests are running the way we want them to.
```
case-eee: webpack is used in a similar way as your asset pipeline in Rails. it helps concatenate, minify, etc. your assets - not your tests

module-4/backend/week_one.md
```diff
@@ -5,18 +5,69 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+
+'this'is the thing that owns the javascript code. This is usually the object when used in the object itself. 
+
 8. What is Webpack and why is it useful?
+
+webpack is a module bundler. It is used in conjunction with selenium to make sure our tests are running the way we want them to. 
+
 9. When/why do you want to use event delegation?
+
+Because you might want something to happen when you clicked a place on the page, you might want something to dynamically move from one page to another.
```
case-eee: not necessarily move dynamically. we want to bind events to elements that are dynamically loaded after the page has already loaded. your JS only runs one time so if an element isn't on the page when your JS runs, then you can't bind an event to that element. this is where event delegation comes into play

module-4/backend/week_one.md
```diff
@@ -5,18 +5,69 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+
+Because you might want something to happen when you clicked a place on the page, you might want something to dynamically move from one page to another. 
+
 10. What is a callback function?
+
+callback functions are ones that wait for something to happen to execute. For instance, this could be making sure that before an object is saved to the database it will make sure it has all the required fields. 
+
 11. What's `npm` and what do we use it for?
 
+npm is the js evivalent of rvm. It starts our server, runs our tests, and installs our packages.
```
case-eee: npm is a package manage. it's the equivalent of Bundler for Ruby. it helps manage the third party packages we leverage in our code

module-4/backend/week_one.md
```diff
@@ -5,18 +5,69 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+npm is the js evivalent of rvm. It starts our server, runs our tests, and installs our packages. 
+
 #### Review  
 12. What's the MVC design pattern? Describe each part of MVC.
+
+Models - where the logic should be and where the classes are. This is where objects are made. 
+Views - What the user sees in a webpage. Should usually have one instance variable and if there are more use a presenter. 
+Controllers - where the direction is taken and things are passed to the view. This is what tells the thing to go where. 
+
+The pattern is that usually a controller is made, passes some sort of thing into the view, and the view and the model work together to make something appear on a webpage.
```
case-eee: the controller should be the part accessing the model and sending it to the view. the view simply renders whatever is provided to it

module-4/backend/week_one.md
```diff
@@ -5,18 +5,69 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+
+Models - where the logic should be and where the classes are. This is where objects are made. 
+Views - What the user sees in a webpage. Should usually have one instance variable and if there are more use a presenter. 
+Controllers - where the direction is taken and things are passed to the view. This is what tells the thing to go where. 
+
+The pattern is that usually a controller is made, passes some sort of thing into the view, and the view and the model work together to make something appear on a webpage. 
+
 13. What are a few ways to optimize a Rails application?
+
+using background workers such as redis, and sidekiq, to make sure the logic is pushed to the models, use presenters to clean up code.
```
case-eee: 👍 you could also cache your code or use AJAX to provide a user with a quicker experience

module-4/backend/week_one.md
```diff
@@ -5,18 +5,69 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
 4. What is the syntax for invoking a function?
+
+function ()
+
 5. What is CORS?
+
+cross origin resource sharing. it is a standard for accessing web resources on different domains. 
+
 6. How do we enable CORS?
+
```
case-eee: how do we enable this in a Rails application specifically though? Or in any other backend application?

PR #422, opened 2017-08-26
CFU Week 2- Sam Nayrouz

snayrouz:
@icorson3 @s-espinosa 

Here you go! 

Got lazy towards the end and didn't draw the schema diagrams for the option questions.

Enjoy the weekend,
Sam

### DIFF
--- a/module-2/backend/week_two.md
+++ b/module-2/backend/week_two.md
@@ -5,6 +5,9 @@ Fork this respository. Answer the questions to the best of your ability. Try to 
 Note: When you're done, submit a PR.
 
 1. At a high level, what is ActiveRecord? What does it do/allow you to do?
+
+ActiveRecord is an object relational mapper(ORM). It allows Ruby objects to map to a database(postgresql in our case). The attributes of that Ruby object become the columns of that specific table in the database.
+
 2. Assume you have the following model:
 
 ```ruby
@@ -13,8 +16,18 @@ end
 ```
 
 What are some methods you can call on `Team`? If these methods aren't defined in the class, how do you have access to them?
+.all
+.find()
+.where()
+.first
+.last
+
+You have access to them because the Team class is inheriting from the ActiveRecord::Base class.
 
 3. Assume that in your database, a team has the following attributes: "id", "name", owner_id". How would you find the name of a team with an id of 4? Assuming your class only included the code from question 2, how could you find the owner of the same team?
+
+Team.find(4).name
+Team.where(owner_id: 4)
 
 4. Assume that you added a line to your `Team` class as follows:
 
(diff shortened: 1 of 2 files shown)


module-2/backend/week_two.md
```diff
@@ -13,9 +16,19 @@ end
...
+.where()
+.first
+.last
+
+You have access to them because the Team class is inheriting from the ActiveRecord::Base class.
 
 3. Assume that in your database, a team has the following attributes: "id", "name", owner_id". How would you find the name of a team with an id of 4? Assuming your class only included the code from question 2, how could you find the owner of the same team?
 
+Team.find(4).name
+Team.where(owner_id: 4)
```
s-espinosa: This will find the Team that has an owner who has an id of 4. We want to find the **owner**.

```ruby
owner_id = Team.find(4).owner_id
Owner.find(owner_id)
```

module-2/backend/week_two.md
```diff
@@ -26,20 +39,82 @@ end
 
 Now how would you find the owner of the team with an id of 4?
 
+@team.owners.where(owner_id: 4)
```
s-espinosa: Go back and review this! Want to have this down cold.

Adding the code above allows us to do the following:

```ruby
Team.find(4).owner
```

module-2/backend/week_two.md
```diff
@@ -26,20 +39,82 @@ end
...
+
 7. Describe the relationship between a foreign key on one table and a primary key on another table.
+
+I believe this answer is similiar to the previous one-> a foriegn key sits on a second table and is the unique id of the first table. That's how the relationship is created.
+
 8. What are the parts of an HTTP response?
 
+- A status line 
+- A status code (100-informational, 200-success, 300-redirection, 400-client error, 500-server error)
+- Header
```
s-espinosa: Status code goes in the status line. HTTP response also has a body.

module-2/backend/week_two.md
```diff
@@ -26,20 +39,82 @@ end
...
+School         Teacher
+id -----       id
+name   |       name
+city   |       grade
+state  |______ school_id
+
+
 6. Give an example of when you might want to store information besides ids on a join table.
+
+To be honest, I don't have one. I thought the purpose of a join table was to normalize and keep the schema and db well designed.
```
s-espinosa: `invoice_items` is a good example: holds quantity and unit price in addition to the foreign keys. Could also store a grade for a class on a join between students and teachers.

module-2/backend/week_two.md
snayrouz: @s-espinosa So I'll be honest, I knew it would return a team with the owner of an id of 4, but the way I understood the question was under the assumption there was only a Team model and not an Owner model. I think the last line of question 3 made me make that assumption.

But either way, good to know.

module-2/backend/week_two.md
snayrouz: Same mistake with the same assumption I made in question 3.

PR #424, opened 2017-08-27
Heft_1706_BEE_Week_2_cfu.md

mikeyduece:
@s-espinosa @icorson3 
Here's my cfu for this past week. Hope you both are having a great weekend. 😃


module-2/backend/week_two.md
```diff
@@ -13,9 +14,12 @@ end
 ```
 
+>`id`, `created_at`, `updated_at`. They are available through the inheritance of the `ActiveRecord::Base`
 
+>`Team.find(4)` `Team.where(owner_id: Team.find(4)).first.owner_id`
```
s-espinosa: For the second part of the question we're looking for an Owner.

```ruby
owner_id = Team.find(4).owner_id
Owner.find(owner_id)
```

module-2/backend/week_two.md
```diff
@@ -26,20 +30,59 @@ end
 
 Now how would you find the owner of the team with an id of 4?
 
+> `Team.first.owner`
```
s-espinosa: This would only work if we had deleted teams 1-3. Need to get the Team with an id of 4 first.

```ruby
Team.find(4).owner
```

module-2/backend/week_two.md
```diff
@@ -26,20 +30,59 @@ end
...
+
 5. In a database that's holding students and teachers, what will be the relationship between students and teachers? Draw the schema diagram.
+
+>Teacher `has_many` students and Student `has_and_belongs_to_many` teachers or just `belongs_to` Teacher depending on the school district I guess.
```
s-espinosa: I like `belongs_to` in this case.

module-2/backend/week_two.md
```diff
@@ -26,20 +30,59 @@ end
...
+
+> foreign key is how one table accesses the relationship to another table. The primary key is the unique identifier of that specific table. Schema is the diagram of how those table are connected, through the foerign keys/ primary keys.
+
 7. Describe the relationship between a foreign key on one table and a primary key on another table.
+
+>it's how the tables access the relationships, ie `has_many`/`belongs_to`
+
 8. What are the parts of an HTTP response?
 
+> The verbs. `GET POST PUT DELETE` The `GET` request responds to a few actions. It will take one from the landing page, the index page, and the show page. For `POST PUT & DELET` there are two steps. First, the `GET` request for the specific `id` and then the corresponding verb for whatever the client has requested. `POST` for new content, `PUT` for updating content and `DELETE` for destroying content
```
s-espinosa: These verbs are part of the **request**.

A response has:

* status line (including a status code)
* headers
* body

module-2/backend/week_two.md
```diff
@@ -26,20 +30,59 @@ end
...
+>`school_id` as a foreign key on the teacher table.
+
 6. Give an example of when you might want to store information besides ids on a join table.
+
+>When that table itself has information that it needs to hold. In the students_teachers example, that joins table could have
+information about what class the teacher is teaching, the max class size allowed, the classroom that it takes place in, it's uartlerly budget...
+
 7. Describe and diagram the relationship between patients and doctors.
+
+>Doctors `has_many` patients and Patients `has_and_belongs_to_many` doctors
```
s-espinosa: For Patients: I would say pick one. Either they only have one doctor or they have many doctors. The `has_and_belongs_to_many` seems to be about creating a join without creating a model for the join table. I generally steer clear.

PR #433, opened 2017-08-27
squee_week_2_CFU

squeemishly:


module-4/backend/week_two.md
```diff
@@ -5,18 +5,83 @@ Fork this respository. Answer the questions to the best of your ability. Try to
 Note: When you're done, submit a PR. 
 
 1. What's one difference between ES5 and ES6?
+```
+ES6 has significantly more syntactic sugar than any of the previous ES's. A few syntactic sugars that ES6 adds:
```
case-eee: 👍

module-4/backend/week_two.md
```diff
@@ -5,18 +5,83 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+```
+synchronous JS runs in the exact order the commands are given
+asynchronous allows you to run processes that take longer outside of the order they are entered. It will start the asynchronous process in the order it was entered, but it will complete the process on the side. When it is done with it's process, it will jump back on the stack to allow any promises it has to complete.
+```
+
 3. What are the four pillars of Object Oriented programming?
+```
+Abstraction - move independent concepts into their own methods. this increases readability and reusability.
+Encapsulation - make methods that can be hidden from public view (e.g. private methods) hidden. this restricts access to essential components of your code that you don't want the average person to have access to.
+Inheritance - reuse as much code as possible. I can use string methods in my "Animal" class without having to write them myself. My code automatically inherits all the methods associated with strings. If I declare a class "Animal" with a method 'walk' and then declare a subclass "Dog", I have access to 'walk' on "Dog".
```
case-eee: great description 🎉

module-4/backend/week_two.md
```diff
@@ -5,18 +5,83 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+```
+Constructor functions allow you to make new classes.
+In ES6, you actually have the ability to declare a 'class'. 
+Hashes are by default JS objects and are treated the same way as classes in Ruby.
+```
+
 5. What are some key concepts to be aware of when refactoring your JavaScript?
+```
+If you refactor out a function from a promise chain, don't include the paranthesis on the called function. If you need to pass in variables, declare an anonymous function and call the new function inside it. There you can pass in variables without immediately invoking the function. 
+Refactoring allows you to better unit test your code. The more you can pull out into individual functions, the easier it is to test and the easier it is to reuse.
```
case-eee: +1 👍

module-4/backend/week_two.md
```diff
@@ -5,18 +5,83 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+
 8. What's the difference between `==` and `===` in JavaScript?
+```
+== 'equals': checks to see if 2 objects equal each other on a simple level. E.G. "2" == 2 //true. "I am squee!" == "I am squee!" //true. true == true //true
+=== 'strict equals': Means the objects must be the same datatype and be exactly the same.  E.G. "2" == 2 //false. "I am squee!" == "I am squee!" //true. true == true //true
+```
+
 9. Why do front end frameworks exist?
+```
+Because jquery is bulky and slow. Much like Rails exists to take the repetitive load off a developer who might otherwise use Sinatra to write an app, React is there to make it easier to dynamically load and change information on the page.
```
case-eee: 👍

module-4/backend/week_two.md
```diff
@@ -5,18 +5,83 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
 #### Review  
 
 10. Why do people say "HTTP is stateless"?
+```
+It doesn't retain information about who you are and what you are doing in any particular app. It continues to occupy the same state before, during, and after your visit to a site. 
+```
+
 11. Describe a RESTful API.
+```
+A RESTful API uses the HTTP verbs in the appropriate ways with the appropriate URLs. It uses a GET request to receive information from the API and the address should look something like /api/v1/thing_i_want. GET should never be used to update, add, or delete information. There are other verbs for that. POST should always be used to make a new item, PUT should be used to update items, and DELETE should be used to remove items.
```
case-eee: great answer!

PR #434, opened 2017-08-28
Mimi Le - CFU - Week 2

mimilettd:
@s-espinosa @icorson3

### DIFF
--- a/module-2/backend/week_two.md
+++ b/module-2/backend/week_two.md
@@ -36,7 +36,7 @@ Now how would you find the owner of the team with an id of 4?
 
 5. In a database that's holding students and teachers, what will be the relationship between students and teachers? Draw the schema diagram.
 
-<img src="http://imgur.com/a/YkBzP>
+<img src="http://i.imgur.com/MYu34Aj.png">
           
 6. Define foreign key, primary key, and schema.
   * A foreign key is a field in one table that refers to the primary key in another table. A primary key is the object's `id`. A schema 
--- a/module-2/backend/week_two.md
+++ b/module-2/backend/week_two.md
@@ -35,6 +35,9 @@ end
 Now how would you find the owner of the team with an id of 4?
 
 5. In a database that's holding students and teachers, what will be the relationship between students and teachers? Draw the schema diagram.
+
+<img src="http://imgur.com/a/YkBzP>
+          
 6. Define foreign key, primary key, and schema.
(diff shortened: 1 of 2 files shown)


module-2/backend/week_two.md
```diff
@@ -26,20 +39,84 @@ end
...
+```
+
 5. In a database that's holding students and teachers, what will be the relationship between students and teachers? Draw the schema diagram.
+
+<img src="http://i.imgur.com/MYu34Aj.png">
+
 6. Define foreign key, primary key, and schema.
+  * A foreign key is a field in one table that refers to the primary key located in another table. A primary key is the object's `id`. A schema shows the current state of the database. When you create a table (`rake db:create migration NAME=<table_name>`) and migrate it (`rake db:migrate`), the schema will update accordingly.
 7. Describe the relationship between a foreign key on one table and a primary key on another table.
+  * A foreign key on one table indicates its' relationship with the primary key on another table (it can be a one-to-many, one-to-one, or many-to-many relationship). Using the example above, a team can belong to an owner and an owner can have many teams. This is a one-to-many relationship. When we specifically indicate this relationship in the `Team` and `Owner` class, we're instructing Ruby to create a foreign key column on the `Team` table using the unique primary key located in the `Owner` table.
```
s-espinosa: Careful here: indicating the relationship on the `Team` and `Owner` class doesn't tell Ruby to create a foreign key column; our migration does that. Indicating the relationship at the class level tells Ruby/ActiveRecord to **look for/use** the foreign key (which exists because of our migrations) to find related rows of data.

module-2/backend/week_two.md
```diff
@@ -26,20 +39,84 @@ end
...
 3. What two columns does `t.timestamps null: false` create in our database?
+  * `created_at` and `updated_at`
 4. In a database that's holding schools and teachers, what will be the relationship between schools and teachers?
+  * one-to-many (a school has_many teachers, a teacher belongs_to a school)
 5. In the same database, what will you need to do to create this relationship (draw a schema diagram)?
+
+<img src="http://i.imgur.com/tgp0Xcv.png">
+
 6. Give an example of when you might want to store information besides ids on a join table.
+  * I don't know. I'm curious, though! What might I want to store aside from ids in a join table?
```
s-espinosa: `invoice_items` from Black Thursday is a good example: it stores quantity and unit_price.

Something like `enrollments` (a join between students and courses) might also be a good place to store something like a grade for the course.

In your example below, we could potentially store `date` and `time` on the `appointments` table.

PR #436, opened 2017-08-28
Update week_two.md

limsammy:


module-2/backend/week_two.md
```diff
@@ -26,11 +36,33 @@ end
...
 
+`Team.find(4).owner`
+
 5. In a database that's holding students and teachers, what will be the relationship between students and teachers? Draw the schema diagram.
+
+Teacher has many Students, a Student belongs to a Teacher
+
+```table students
+|      id    |
+|     name   |
```
icorson3: a belongs to should have a relationship on that table. A student will have a teacher_id on their table, not the way you have it

module-2/backend/week_two.md
```diff
@@ -26,11 +36,33 @@ end
...
+|    name    |
+| student_id |```
+
 6. Define foreign key, primary key, and schema.
+
+A primary key is the `id` of a table. It is the key that is inserted to an associated table. The key referenced in the associated table is the `foreign key`. The schema is a blueprint of the database and all of your migrations.
+
 7. Describe the relationship between a foreign key on one table and a primary key on another table.
+
+A foreign key is the has_many part of an association, while a primary key is the id being inserted as the foreign key.
```
icorson3: a foreign key on one table will refer to the primary key of the table it is references

icorson3:
see comments. nice job 🎉

PR #437, opened 2017-08-28
Danny Trujillo - Week 2 CFU

djtrujillo:


module-2/backend/week_two.md
```diff
@@ -13,8 +14,11 @@ end
 ```
 
+    * The inheritance from ActiveRecord allows Team to call active record methods such as Where and Find. Since they aren't defined in the class they are inherited from the ActiveRecord::Base command. 
 
+    * team = Team.find(4)
```
icorson3: Team.find(4).name

icorson3:
see comments. nice job 🎉

PR #452, opened 2017-09-01
Kali Borkoski  CFU week3 

JunePaloma:
@s-espinosa 
@icorson3 
Thanks!!


module-2/backend/week_three.md
```diff
@@ -10,14 +10,65 @@ Note: When you're done, submit a PR with a reflection in the comments about how
...
+<h1>Add a new horse for to <%= @owner.name %>'s collection</h1>
+<%= form_for ([@owner, @horse]) do |f| %>
+<%= f.label :name %>
+<%= f.textfield :name %>
+<%= f.label :winnings %>
+<%= f.label :textfield %>
+<%= f.sbumit %>
+<% end %>
+
+I changed this a little because I don't have a great understanding of how @owner is working here and I'd love some illumination; it's the @owner we defined in our controller -- which has a specific id, which in turn allows us to associate the horse with that spefific owner. But I don't understand why/how.
```
s-espinosa: The `form_for` uses the arguments to determine the action of the form (this is the route that the form will send its request to). The form above implies that we want our users to be able to create new horses that belong to users. It also implies to me that we want users to be able to create these horses by visiting a different page for each owner (as opposed to selecting an owner from a dropdown or something like that).

We need `@owner` to be passed as part of the argument because the `form_for` is going to use these arguments to create an action like `/owners/4/horses`. The ID in that URI refers to the owner's ID. Without passing `@owner` as an argument, the `form_for` wouldn't know the ID of the owner, and because of that wouldn't be able to generate the `/owners/4/horses` route as an action for the form.

On the other hand, the `form_for` also needs `@horse` as an argument. It uses `@horse` for two reasons: 1) to generate the action on the form (the `/horses` piece of it), and 2) to do some error checking on the contents of the form to make sure that we are only asking users to input information that corresponds to attributes of a Horse as defined in our DB.

PR #469, opened 2017-09-05
Mimi Le - Week 3 - CFU

mimilettd:
@s-espinosa  @icorson3

### DIFF
--- a/module-2/backend/week_three.md
+++ b/module-2/backend/week_three.md
@@ -10,14 +10,60 @@ Note: When you're done, submit a PR with a reflection in the comments about how 
 ### Questions
 
 1. What is the entry at the command line to create a new rails app?
+
+```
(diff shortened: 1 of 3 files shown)


module-2/backend/week_three.md
```diff
@@ -10,14 +10,60 @@ Note: When you're done, submit a PR with a reflection in the comments about how
...
+  <%= f.label :name %>
+  <%= f.text_field :name %>
+  <%= f.label :age %>
+  <%= f.text_field :age %>
+  <%= f.submit %>
+<% end %>
+```
+
 12. Why do we want to validate our models?
+  * To ensure that only valid data is saved into our database. Say for example we want every user to input a valid email address. Rails will not allow the user to create an account unless an e-mail is entered by doing `validates :email, presence: true`.
```
icorson3: also uniqueness!

icorson3:
nice work @mimilettd 🎉

PR #478, opened 2017-09-05
Update week_three.md

limsammy:
This went about 7/10 for me.

### DIFF
--- a/module-2/backend/week_three.md
+++ b/module-2/backend/week_three.md
@@ -10,14 +10,54 @@ Note: When you're done, submit a PR with a reflection in the comments about how 
 ### Questions
 
 1. What is the entry at the command line to create a new rails app?
+
+rails new NAME -T --database=postgresql --skip-spring --skip-turbolinks
+
 2. What do Models generally inherit from in rails?
+
+Rails' ApplicationRecord
+
 3. What do Controllers generally inherit from in a rails project?
+
+Rails' ApplicationController
+
 4. How would I create a route if I wanted to see a specific horse in my routes fitle assuming I'm sticking to standard conventions and that I didn't want other CRUD functionality?
+
+resources :horse, only [:show]
+
 5. What rake task is useful when looking at routes, and what information does it give you?
+
+rake routes, this gives you the action, path, and URL
+
 6. What is an example of a route helper? When would you use them?
+
+A route helper is something like `link_to` where you can pass it ActiveRecord objects.
+
 7. What's the difference between what `_url` and `_path` return when combined with a routes prefix?
+
+\_url is the URI pattern, whereas path is the actual path to the object/view
+
 8. What are strong params and why are the necessary?
+
+Strong params are calling params directly, not through your protected validation method. This is necessary when allowing the user to sort.
+
 9. What role does `form_for` play in helping us create our forms?
+
+`form_for` allows us to pass in a ruby object (or multiple objects) for easy form building using erb tags.
+
 10. How does `form_for` know where to submit the user's input?
+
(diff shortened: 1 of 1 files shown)


module-2/backend/week_three.md
```diff
@@ -10,14 +10,54 @@ Note: When you're done, submit a PR with a reflection in the comments about how
...
+
+resources :horse, only [:show]
+
 5. What rake task is useful when looking at routes, and what information does it give you?
+
+rake routes, this gives you the action, path, and URL
+
 6. What is an example of a route helper? When would you use them?
+
+A route helper is something like `link_to` where you can pass it ActiveRecord objects.
```
icorson3: horses_path is a route helper and i would use it in the view/controller/test as a reference to a path

module-2/backend/week_three.md
```diff
@@ -10,14 +10,54 @@ Note: When you're done, submit a PR with a reflection in the comments about how
...
+
 7. What's the difference between what `_url` and `_path` return when combined with a routes prefix?
+
+\_url is the URI pattern, whereas path is the actual path to the object/view
```
icorson3: _path helpers provide a site-root-relative path. You should probably use this most of the time. i.e. "/horses" Internal to my app
_url helpers provide an absolute path, including protocol and server name. i.e "localhost:3000/horses" Accessing from browser

module-2/backend/week_three.md
```diff
@@ -10,14 +10,54 @@ Note: When you're done, submit a PR with a reflection in the comments about how
...
+
 8. What are strong params and why are the necessary?
+
+Strong params are calling params directly, not through your protected validation method. This is necessary when allowing the user to sort.
+
```
icorson3: strong params are actually a private method typically that we tell our app what attributes we will allow for each object created and if there are attributes outside of those params, they are not allowed on the object.

module-2/backend/week_three.md
```diff
@@ -10,14 +10,54 @@ Note: When you're done, submit a PR with a reflection in the comments about how
...
 9. What role does `form_for` play in helping us create our forms?
+
+`form_for` allows us to pass in a ruby object (or multiple objects) for easy form building using erb tags.
```
icorson3: sets up a html form

module-2/backend/week_three.md
```diff
@@ -10,14 +10,54 @@ Note: When you're done, submit a PR with a reflection in the comments about how
...
+
 10. How does `form_for` know where to submit the user's input?
+
+Through it's initialization, we pass it ruby object(s).
```
icorson3: the instance variable references a objects path and the new or edit html provides us with which verb we are using

module-2/backend/week_three.md
```diff
@@ -10,14 +10,54 @@ Note: When you're done, submit a PR with a reflection in the comments about how
...
+
 11. Create a form using a `form_for` helper to create a new `Horse`. 
+
+```form_for @horse do |f|
+    f.label :name
```
icorson3: provide erb tags

module-2/backend/week_three.md
```diff
@@ -10,14 +10,54 @@ Note: When you're done, submit a PR with a reflection in the comments about how
...
+    f.text_field :name
+    
+    f.submit
+ end```
+
 12. Why do we want to validate our models?
+
+So dangerous paramaters are not passed to our controller and database (prevents attacks like SQLi)
```
icorson3: so data is complete and unique

icorson3:
see comments @limsammy 🎉

PR #495, opened 2017-09-11
Randy Springer 1706 BE CFU Week 4

randy-darsh:


module-2/backend/week_four.md
```diff
@@ -10,14 +10,49 @@ Note: When you're done, submit a PR with a reflection in the comments about how
 ### Questions
 
 * What is a cookie?
+
```
icorson3: a small piece of data sent from a website and stored on the user's computer by the user's web browser while the user is browsing.

module-2/backend/week_four.md
```diff
@@ -10,14 +10,49 @@ Note: When you're done, submit a PR with a reflection in the comments about how
...
+stored data from a web site like what's in your shopping cart or links you have clicked on
+
 * What’s the difference between a session and a cookie?
+
+a session in your log in and specific to your user. a cookie is saved data outside of the session
```
icorson3: Sessions allow you to store away individual pieces of information just like with cookies, but the data gets stored on the server instead of the client

module-2/backend/week_four.md
```diff
@@ -10,14 +10,49 @@ Note: When you're done, submit a PR with a reflection in the comments about how
...
+
 * What’s a flash and when do you want to use flashes?
+
+a flash is how you can display a temporary message like: 'successful login', 'logged out', etc.
+
 * Why do people say “HTTP is stateless”?
+
+because the cookies and sessions is what hold state on your computer
```
icorson3: not entirely...Because a stateless protocol does not require the server to retain session information or status about each communications partner for the duration of multiple requests. HTTP is a stateless protocol, which means that the connection between the browser and the server is lost once the transaction ends

module-2/backend/week_four.md
```diff
@@ -10,14 +10,49 @@ Note: When you're done, submit a PR with a reflection in the comments about how
...
+
 * What’s authentication? Explain.
+
+authentication is how you can tell a user is who they say they are. i.e. email, username, password
+
 * What’s the difference between authentication and authorization?
+
+authorization is the power to do something you set. for example, if you dont want users to be able to delete gifs then you would create an admin role and only give the admin the ability to do it.
```
icorson3: authentication = i am who i say i am vs authorization = i have permission to be here

module-2/backend/week_four.md
```diff
@@ -10,14 +10,49 @@ Note: When you're done, submit a PR with a reflection in the comments about how
...
+
 * What’s a before filter?
+
+a method to check if something is true before it executes it. like redirecting to a 404 page if the current user is not an admin. like an extra layer of security so even if they do go to the specific route you may have only set up for admin, they will not be be to access it or even really know if it's there.
```
icorson3: it is a callback so it is not checking if it is true but it is performation a function BEFORE some other function or method

module-2/backend/week_four.md
```diff
@@ -10,14 +10,49 @@ Note: When you're done, submit a PR with a reflection in the comments about how
...
+
 * How do we keep track of a user once they’ve logged in?
+
+with sessions
+
 * When do you want to namespace a resource? When do you want to nest a resource? What's the differences between those two approaches?
+
```
icorson3: You would want to namespace a resource when you want to grant authorization to a group of users (for example: admins) to do certain things (like `delete` or `create` objects/rows).
You would want to nest a resource when there is a relationship between the parent resource and child resource (the URI will typically look like this: `/users/1/gifs`)

module-2/backend/week_four.md
```diff
@@ -10,14 +10,49 @@ Note: When you're done, submit a PR with a reflection in the comments about how
...
+nesting is better if it doesnt matter who can access something. if you namespace it drops the id from the route so it allows for creating a 
+
 * At a high level, what tools can you use to implement authorization? How would you use them?
+
+namespaceing, roles and the before filter. namespace admin into whatever actions you want to be authorized to only them. use roles to make sure there are at least 2 types of users and set one to admin. give that account access to whatever you want and use the before filter to make sure only they can see it.
+
 * What's an enum, and what advantages does it offer? What data type needs to be in your database to use an enum? Where do you declare an enum?
+
+enum is how you can allow for an admin and user role by an integer. you can default it to zero so normal users wont even know it exists. then set yourself to one so you can do all the admin things you set up for yourself. datatype is an integer and you declare them in the migration and the model.
```
icorson3: not ALWAYS a role and have the flexibility of a string, but the speed and efficiency of an integer

module-2/backend/week_four.md
```diff
@@ -10,14 +10,49 @@ Note: When you're done, submit a PR with a reflection in the comments about how
...
+
 * What are some strategies you can use to keep your views DRY?
+
```
icorson3: not quite. partials, shared controller methods, presenters

icorson3:
please see comments. a few things could use a second look @randy-darsh

PR #506, opened 2017-09-11
Week 4 - Dan P 

dphilla:


module-2/backend/week_four.md
```diff
@@ -10,14 +10,64 @@ Note: When you're done, submit a PR with a reflection in the comments about how
 ### Questions
 
 * What is a cookie?
+
+  A cookie is a piece of data assigned to and saved in your browser when you visit a website
+
 * What’s the difference between a session and a cookie?
+
+  a session is encrypted, and can hold more data(?)
```
s-espinosa: Mostly the encrypted part. Intrigued by more data? I haven't heard that, but could be something I just haven't heard yet.

module-2/backend/week_four.md
```diff
@@ -10,14 +10,64 @@ Note: When you're done, submit a PR with a reflection in the comments about how
...
+  a flash is a hash that can be used to send temporary messages to a given view
+
 * Why do people say “HTTP is stateless”?
+
+  Because each response from a server doesn't carry 'state' from other responses
+
 * What’s authentication? Explain.
+
+   Authentication is what users to have unique experiences on a web app, e.g.
+   a logged-in has a more options than a non logged-in user.
```
s-espinosa: I would say the idea of `having more options` is probably more closely related to authorization. Authentication is primarily about knowing who a user is, not necessarily what they can do (though the two are tightly coupled).

module-2/backend/week_four.md
```diff
@@ -10,14 +10,64 @@ Note: When you're done, submit a PR with a reflection in the comments about how
...
+
 * What’s the difference between authentication and authorization?
+
+  Authorization is allows for different types of user roles and is typically
+  achieved through namespacing different user functions through different
+  controllers.
+  Authentication verifies that a given user interacting with a site is "who
+  they say they are" and have access to records correspoding with their account.
```
s-espinosa: 👍

module-2/backend/week_four.md
```diff
@@ -10,14 +10,64 @@ Note: When you're done, submit a PR with a reflection in the comments about how
...
+  a namesspaced resources is not reliant on a specific id for a user when accessing a URI,
+  but rather on their authorization status. a nested resources will rely on the specific
+  parent resources and whatever child resources are spelled-out in the URI.
+
 * At a high level, what tools can you use to implement authorization? How would you use them?
+
+  enums, usually on the resource model, allow for different roles to be managed
+  with associated methods based on the naming of those roles. With this, we can
+  write a method on the application controller that can be available to all controllers
+  for checking a user's role before a given action.
```
s-espinosa: Yup! That's how we've used it up to this point. More generally it can be used for any attribute that we know will only have a limited number of options that we want to store in the database as a number and refer to in our code using a word. Also gives us some nice `?` and `!` methods.

Other tools that we're using for authorization: before filter, inheritance.

PR #538, opened 2017-10-06
Aaron Schreck 1708-B

aschreck:


module-2/backend/week_one.md
```diff
@@ -5,28 +5,57 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+  Patch: Patch updates part of a resource, but not the whole thing. 
+  Put: updates a whole resource. 
 2. What is Sinatra?
+  Sinatra is a framework for making applications that incorporates Active Record and SQL databases. 
 4. What is MVC?
+  MVC stands for the Mode-View-Controller model for structuring applications
 5. Why do we follow conventions when creating our actions/path names in our Sinatra routes?
+  We use the RESTful route structure so that our applications are easy to read and understand according to a standardized structure.  
 6. What types of variables are accessible in our view templates without explicitly passing them?
+  Global Variables
```
icorson3: instance variables (vs. local) was what we were looking for but global works too, i guess!

module-2/backend/week_one.md
```diff
@@ -5,28 +5,57 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
+    @name = "Mr. Ed"
     erb :index
   end
   ```
 
 9. What's the purpose of ERB?
+  Embedded Ruby files provide a place to integrate HTML and Ruby into one application view. 
+  So that you can a variety of tests without affecting the database you are using for production.
```
icorson3: the databases are development and test...a production database should be separate as well but would be dealt with on a different level

module-2/backend/week_one.md
```diff
@@ -5,28 +5,57 @@ Fork this respository. Answer the questions to the best of your ability. Try to
...
 9. What's the purpose of ERB?
+  Embedded Ruby files provide a place to integrate HTML and Ruby into one application view. 
 10. Why do I need a development AND test database?
+  So that you can a variety of tests without affecting the database you are using for production. 
 11. What is CRUD and why is it important?
+Create, Read, Update, Delete. These four things constitute the essential/basic function of a web app.
 12. What does HTTP stand for? 
+HyperText Transfer Protocol
 13. What are the two ways to interpolate Ruby in an ERB view template? What's the difference between these two ways?
+<%= %> or #{}
```
icorson3: looking for `<&= %>` or `<% %>`. one shows the information, the other allows us access to ruby

icorson3:
nice work, a few comments @aschreck 🎉

PR #565, opened 2017-10-09
1708 BE Young Jung 

seoulstice:


module-2/backend/week_one.md
```diff
@@ -1,32 +1,96 @@
...
+
+  get '/restaurants/edit' --> erb :edit
+  put '/restaurants/:id' --> redirect "/restaurants/#{id}"
+  To edit a restaurant record
+
+  delete '/restaurants/:id' --> redirect '/restaurants'
+  To delete a restaurant record
+
+16. What's a migration?
+  A migration is transferring data from one source to another.  In our case so far, from a CSV
```
s-espinosa: Migrations are directions for how to create the structure of your database. In order to actually change the structure of the database you need to run the migration.

Seeding, on the other hand, populates your database.

module-2/backend/week_one.md
```diff
@@ -1,32 +1,96 @@
...
+  to an online database.
+
+17. When you create a migration, does it automatically modify your database?
+
+  No, you have to seed the database as well.
```
s-espinosa: See above.

module-2/backend/week_one.md
```diff
@@ -1,32 +1,96 @@
...
+
+18. How does a model relate to a database?
+  The model determines how/what columns are created on a database.  The model also contains methods used
```
s-espinosa: A migration determines the columns that are created in the database. The model provides us a way to interact with the database in Ruby.

Generally, a model will represent a table in the database. An instance of that model will correspond to a particular row.

module-2/backend/week_one.md
```diff
@@ -1,32 +1,96 @@
...
+  for extracting data from database to be use on the view.
+
+19. What is the difference between `#new` and `#create`?
+  Create makes a new instance and also deletes that instance once it is done being used
```
s-espinosa: `.new` makes the object in memory, but doesn't save it to the database until you run `.save` on the new instance. `.create` does both together.

PR #571, opened 2017-10-09
answer mod2 week1 CFU

JosiMcClellan:


module-2/backend/week_one.md
```diff
@@ -1,32 +1,105 @@
...
+get '/horses' do
+  @count = 1
+  erb :index
+end
+```
+
+##### How would I pass a local variable `name` with a value of `Mr. Ed` to the view?
+```ruby
+get '/horses' do
+  erb :index, locals => { name: 'Mr. Ed' }
```
s-espinosa: Interesting! This should work. I think with the question saying **local variable** I would expect to save 'Mr. Ed' to a variable first and then pass that variable.

module-2/backend/week_one.md
```diff
@@ -1,32 +1,105 @@
...
+end
+```
+
+##### What's the purpose of ERB?
+To insert dynamic content, using ruby, into HTML
+
+##### Why do I need a development AND test database?
+The test database tends to be cleared after every test.
```
s-espinosa: Also, I don't want to clog my dev database with test data.

module-2/backend/week_one.md
```diff
@@ -1,32 +1,105 @@
...
+
+##### What is CRUD and why is it important?
+**C** reate
+**R** ead
+**U** pdate
+**D** estroy
+It's how you database.
```
s-espinosa: Ha!

PR #578, opened 2017-10-15
Emily Wise 1708-BE- Week 2

wiseemily88:


module-2/backend/week_two.md
```diff
@@ -13,9 +15,16 @@ end
 ```
 
+.join, .average, .find, .create, .update, .maximum, .minimum, .group, .pluck
+We can call these methods because we are inheriting from the ActiveRecord Base model.
 
+team_4 = Team.find(4)
+Owner_id = team_4.owner_id << and then I believe you would need to create a table for owners and a reference to Team so we could find out the name of the owner- right now there is not connection so you can not find the name from team. you would need to do something like this
```
s-espinosa: I would change `Owner_id` to `owner_id`, but otherwise this looks good.

module-2/backend/week_two.md
```diff
@@ -13,9 +15,16 @@ end
...
+
+Owner_name = Owner.find(owner_id)
```
s-espinosa: Same thing here with `owner_name`.

module-2/backend/week_two.md
```diff
@@ -26,20 +35,110 @@ end
...
 2. Name your three favorite ActiveRecord rake tasks and describe what they do.
+rake db:drop - drops the database
+rake db:create - creates the datbase
+rake db: migrate - sets the migrations
+
 3. What two columns does `t.timestamps null: false` create in our database?
+created_at and updated_at
+
 4. In a database that's holding schools and teachers, what will be the relationship between schools and teachers?
+Many to many
```
s-espinosa: Interesting! Would a teacher have many schools? I can see a case where they would if they taught different places, but generally I think of teachers only belonging to a single school.

PR #594, opened 2017-10-16
Lauren 1708

BLaurenB:


module-2/backend/week_two.md
```diff
@@ -1,35 +1,64 @@
...
 class Team << ActiveRecord::Base
 end
 ```
-
+You could call .all, .count, .find, .find_by, .where, which are all methods defined by ActiveRecord.
 
+Team.find(4).name
+Team.find(4).owner
```
s-espinosa: Without the `belongs_to` you would need to be a little creative to find the owner.

```
team = Team.find(4)
Owner.find(team.owner_id)
```

module-2/backend/week_two.md
```diff
@@ -1,35 +1,64 @@
...
 4. Assume that you added a line to your `Team` class as follows:
-
 ```ruby
 class Team << ActiveRecord::Base
   belongs_to :owner
 end
 ```
-
+Team.find(4).owners
```
s-espinosa: Should still be singular `owner`.

PR #681, opened 2017-12-03
Heft_1706BEE_week_one_CFU

mikeyduece:
@laurenfazah @kat3kasper

### DIFF
--- a/module-4/backend/week_one.md
+++ b/module-4/backend/week_one.md
@@ -1,21 +1,71 @@
 ## Week One - Module 4 Recap
 
-Fork this respository. Answer the questions to the best of your ability. Try to answer them with limited amount of external research. These questions cover the majority of what we've learned this week. 
+Fork this respository. Answer the questions to the best of your ability. Try to answer them with limited amount of external research. These questions cover the majority of what we've learned this week.
 
-Note: When you're done, submit a PR. 
+Note: When you're done, submit a PR.
 
 1. What's the most useful thing you learned from completing the intermission week work?
+
+>That I can make the connection between Ruby and JS and apply those changes pretty well.
+
 2. What are some tools to help debug JavaScript code?
+
+>Debugger, console.log(), breakpoints
+
 3. What are some tools you need in order to unit test your JavaScript?
+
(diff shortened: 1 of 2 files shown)


module-4/backend/week_one.md
```diff
@@ -1,21 +1,71 @@
...
+
 3. What are some tools you need in order to unit test your JavaScript?
+
+>Mocha, Chai, npm, node modules
+
 4. What is the syntax for invoking a function?
+
+>functionName()
+
 5. What is CORS?
```
kat3kasper: 5 & 6 are questions from an old version of the repo. Be sure to pull the latest version

module-4/backend/week_one.md
```diff
@@ -1,21 +1,71 @@
...
+
+>Similar to ruby self. it refers to the current spot in the code that it is referring to in the scope fo the function.
+
 8. What is Webpack and why is it useful?
+
+>A module bundler. Its main purpose is to bundle JavaScript files for usage in a browser, yet it is also capable of transforming, bundling, or packaging just about any resource or asset.
+
 9. When/why do you want to use event delegation?
+
+>When you may have multiple 'buttons' that require an event listener. This way
```
kat3kasper: Also when you may want to listen to an event that doesn't yet exist on the page. If you place an event on the parent element, event bubbling will cause the parent to catch events that happen on any children that are currently there or will be there in the future.

module-4/backend/week_one.md
```diff
@@ -1,21 +1,71 @@
...
+then calls the model to get any information required from that request. Once that
+information is gathered, it sends that to the view to build and populate the html.
+The view then sends that back to the controller which then sends all of that back to
+the router in an http response.
 12. What are a few ways to optimize a Rails application?
+
+> Caching, eager loading, background workers, chron jobs...
+
 13. What's a background worker? When would we want to use a background worker?
+
```
kat3kasper: Yes, I would like this to be a bit more technical though. You give an example a bit on why its used but not very clear on what it is. Background workers are tasks that can be run asynchronously to allow you code to continue running while you process these potentially long running tasks separately.

PR #696, opened 2017-12-04
Joan Harrington Mod 4/Week 1

sidewinder2020:


module-4/backend/week_one.md
```diff
@@ -1,21 +1,49 @@
...
+dev tools: source code/debugger/console/element/network
+
 3. What are some tools you need in order to unit test your JavaScript?
+mocha/chai
+
 4. What is the syntax for invoking a function?
+functionName() - the parentheses at the end
+
 5. What is CORS?
+Cross Origin Resource Sharing
```
kat3kasper: Questions 5 & 6 are from an old version of the repo, please pull the latest so you will have the updated questions in the future

module-4/backend/week_one.md
```diff
@@ -1,21 +1,49 @@
...
+add an HTTP headers tag? I still need to enable - haven't pushed to production
+
 7. What's `this` in JavaScript?
+the entirety of whatever context you're working in - and all of the prototypes accessible to you for that context
+
 8. What is Webpack and why is it useful?
+a module bundler - it bundles all of the 'make it pretty' files in order to export them to your browser (it's a wannabe asset pipeline)
+
 9. When/why do you want to use event delegation?
+when we set up an event listener to a listen/react to a particular event
```
kat3kasper: This describes more of what an event listener is than event delegation. Event delegation allows us to take advantage of event bubbling to handle events that may happen on elements that currently exist in the dom or may be added in the future. We can add an event to a parent element and event bubbling will always cause our event to be triggered whether the child element existed on page load or not.

module-4/backend/week_one.md
```diff
@@ -1,21 +1,49 @@
...
+Model View Controller
+M - relationships, queries on a class - talks to database
+V - presentation of said data
+C - grabs info from the model, and tailors it for the view
+
 12. What are a few ways to optimize a Rails application?
+tailoring how you store things in your database (including storing strings vs integers/indexing/etc.), and caching
+
 13. What's a background worker? When would we want to use a background worker?
+handles any delegated operations in the background (that we don't want taking up resources) - sidekiq was the one we used in mod 3
```
kat3kasper: When would we use background workers? Sending an email, processing an image, etc

PR #814, opened 2018-01-29
Miguel Silva CFU Week 1

MiguelSilva1997:


module-4/backend/week_one.md
```diff
@@ -1,19 +1,50 @@
...
-Fork this respository. Answer the questions to the best of your ability. Try to answer them with limited amount of external research. These questions cover the majority of what we've learned this week. 
+1. What's the most useful thing you learned from completing the intermission week work?
 
-Note: When you're done, submit a PR. 
+  The most important thing that I learned from completing the intermission week work is JavaScript syntax.
 
-1. What's the most useful thing you learned from completing the intermission week work?
 2. What are some tools to help debug JavaScript code?
+
+  Some tools to help me debug JavaScript are pry, console.log, and the console in your browser.
```
kat3kasper: Yup! Along with some of the developer tools provided by chrome: elements panel, source, network, etc

module-4/backend/week_one.md
```diff
@@ -1,19 +1,50 @@
...
+
+  function name(){
+
+  }
+
+  name()
+
 5. What's `this` in JavaScript?
+
+  `this` is the outer object of the current object that you are in.
```
kat3kasper: Could you give a little more detail here? What is `this` when you are inside an object, function, or global?

module-4/backend/week_one.md
```diff
@@ -1,19 +1,50 @@
...
+
 6. What is Webpack and why is it useful?
+
+  Webpack is a bundler or boilerplate for JavaScript to help you build a file blueprint faster.
+
 7. When/why do you want to use event delegation?
+
+  You want to use event delegation when you want to control what event listeners get activated when the user performs an action.
```
kat3kasper: I would say that is more of a description of using event handlers. Event delegation is specific to using event bubbling to trigger events for elements in the DOM that may or may not currently exist on the page

module-4/backend/week_one.md
```diff
@@ -1,19 +1,50 @@
...
+
+  name()
+
 5. What's `this` in JavaScript?
+
+  `this` is the outer object of the current object that you are in.
+
 6. What is Webpack and why is it useful?
+
+  Webpack is a bundler or boilerplate for JavaScript to help you build a file blueprint faster.
```
kat3kasper: It is not only used to build a file blueprint, you can use transpilers such as babel to compile ES6 code to ES5, you can build all JS files into 1 file, as well as minifying this file so that its smaller and faster for browsers

PR #954, opened 2018-05-06
Jamison Ordway BE2, W1 CFU

jamisonordway:


jamisonordway:
@iandouglas @dionew1  I think that my answers from week two are included in this PR. Please let me know if there are any problems with it

module-2/backend/week_two.md
```diff
@@ -1,24 +1,34 @@
...
 ```
 
+
+Find, group, joins are all ActiveRecord methods which the above Team class inherits.
+
 
+Team.find(4), Team.find(4).owner_id
```
iandouglas: team = Team.find(4)
owner = Owner.find(team.owner_id)

module-2/backend/week_two.md
```diff
@@ -29,32 +39,89 @@ end
...
+- Body
+
 
 ### Optional Questions
 
 1. Name your five favorite ActiveRecord methods (i.e. methods your models inherit from ActiveRecord) and describe what they do.
+
+sum (adds numbers together)
+joins (for join tables)
+find_by (allows to search by attribute)
```
iandouglas: .find_by_attribute(value)
Contact.find_by_first_name('bob')
Contact.find_by(first_name: 'bob')

PR #956, opened 2018-05-07
Justine Troyke - Mod 2: Week 5 - CFU 

justinetroyke:


module-2/backend/week_two.md
```diff
@@ -16,8 +18,15 @@ end
 ```
 
+  Team.first
+  Team.all
+  Team.create
+  Team.validates
 
+  1. Team.find(4)
```
iandouglas: Team.find(4).name

module-2/backend/week_two.md
```diff
@@ -16,8 +18,15 @@ end
...
+  2. id = Team.find(4).owner_id
+    Owner.find(id).name
```
iandouglas: team = Team.find(4)
Owner.find_by(team.owner_id)

PR #961, opened 2018-05-07
Zachary Thomas check for understanding

zdcthomas:


zdcthomas:
Zachary Thomas Mod 2 Week 2 CFU

module-2/backend/week_two.md
```diff
@@ -16,8 +18,11 @@ end
 ```
 
+* .all, .first, .last, .(any other column name in the associated table). class Team inherits from Base in the active record module, which provides Team with the base active record methods.
 
+* Team.find(4).name
+* Owner.find(Team.find(4))
```
iandouglas: you would need an integer to call `Owner.find(something)`, and `Team.find(4)` will return a Team object. What else would you need here?

PR #962, opened 2018-05-07
Tyler Westlie Week 3 CFU

tywestlie:


module-2/backend/week_two.md
```diff
@@ -29,32 +34,74 @@ end
 
 Now how would you find the owner of the team with an id of 4?
 
+* team_owner = Team.find_by(owner_id: 4)
```
iandouglas: this does the same thing as your previous code, just more compact. What we were looking for was something like `owner = Team.find(4).owner` or something like that.

